PR #14, opened 2019-04-08
Updated pom with Wiki link and scm tag

nbhoski:
Post Jenkins fork as per instruction on the hosting issue updated the Wiki link in pom and also added <scm> tag details.


nbhoski:
Creation of wiki is not mandatory, However it is recommended to have it for easy web searches. I have created Wiki with mandatory fields required with minimum information. please review the same [here](https://wiki.jenkins.io/display/JENKINS/MATLAB+Plugin). We can keep updating this as and when required .

PR #18, opened 2019-04-26
Fix to supoprt Jenkins environment variables.

nbhoski:
Added fix to support Jenkins environment variables during build.

### DIFF
--- a/src/main/java/com/mathworks/ci/MatlabBuilder.java
+++ b/src/main/java/com/mathworks/ci/MatlabBuilder.java
@@ -394,10 +394,11 @@ public class MatlabBuilder extends Builder implements SimpleBuildStep {
             if (rel.verLessThan(BASE_MATLAB_VERSION_BATCH_SUPPORT)) {
                 ListenerLogDecorator outStream = new ListenerLogDecorator(listener);
                 return launcher.launch().pwd(workspace)
+                        .envs(env)
                         .cmds(constructDefaultMatlabCommand(isLinuxLauncher)).stderr(outStream)
                         .join();
             } else {
-                return launcher.launch().pwd(workspace).cmds(constructMatlabCommandWithBatch())
+                return launcher.launch().pwd(workspace).envs(env).cmds(constructMatlabCommandWithBatch())
                         .stdout(listener).join();
             }
         } catch (MatlabVersionNotFoundException e) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -394,10 +394,11 @@ private int execMatlabCommand(Run<?, ?> build, FilePath workspace, Launcher laun
             if (rel.verLessThan(BASE_MATLAB_VERSION_BATCH_SUPPORT)) {
```
jwilkick-mw: I think this could be a bit cleaner if you only put what is different between each version in the if blocks and keep the common stuff outside. For example:

```suggestion
ProcStarter matlabLauncher = launcher.launch().pwd(workspace).envs(env);

if(rel.verLessThan(BASE_MATLAB_VERSION_BATCH_SUPPORT){
     matlabLauncher = matlabLauncher.cmds(...).otherstuff(...);
}else{
    // ... different flow 
}

matlabLauncher.join();
```
I also noticed that one side of this if statement sets a stderr handler and the other side sets a stdout handler. As written, it seems like the plugin will only handle stdout or stderror depending on the version of MATLAB. Is that intentional?  Or should we always be capturing both and handling them, just with different strategies depending the MATLAB version?

I can also see this if statement picking up more cases as we start accounting for more differences between MATLAB versions.  I think there might be an opportunity for using the Factory pattern and polymorphism here to manage that complexity as it grows.

nbhoski: Suggestion for re writing if with specific blocks is implemented.

The Std error and stout is intentional here. The difference lies in the way MATLAB handles this for -batch and for -r option. As per Alex Long who developed -batch. MATLAB redirects stdout and dtderr to stderr stream when -r option with -log is used. However -batch is automatically handles this by redirecting it to stdout . 

Alex Long's email :

> _Hi Nikhil,_
_Using the -log flag with -wait does 2 things. It causes the stdout and stderr streams of the MATLAB starter executable to be shared with the MATLAB main executable and it redirects stdout to stderr. Because stdout is being mirrored in stderr, you see duplicates in the Command Prompt unless you use > NUL to tell stdout to be discarded._

> _I'm guessing this works differently in your Java console because it uses different syntax or > options for text redirection. Based on the terminal prompt, it looks like the Java console is a > bash shell so maybe try using_
> /dev/null
instead of
> NUL
since NUL is unique to the Windows Command Prompt (cmd.exe).

> **From 18b onwards, I'd suggest using -batch which should handle console output correctly on > all platforms without needing any extra command line switches or_**

nbhoski: Factory method for If statement is good suggestion. We are already planning to use factory pattern for writing our CI scratch file. i will include this point when i table the RF and review it. i have created a Jira plan for the same. https://jira.mathworks.com/browse/CI-245

PR #19, opened 2019-04-30
Added unit tests to improve code coverage

nbhoski:


src/test/java/com/mathworks/ci/MatlabBuilderTest.java
```diff
@@ -279,4 +282,134 @@ public void verifyCustomCommandInvokedForBatchMode() throws Exception {
...
+        Assert.assertFalse(runOption.getBooleanByName("fakeCommand"));
+    }
+    
+    /*
+     * Test to verify when MATLAB version is older the R2017a
+     */
+
+    @Test
+    public void verifyMatlabVersionOlderThanR17a() throws Exception {
+        this.matlabBuilder.setLocalMatlab(getMatlabroot("R2018b"));
```
jwilkick-mw: Should the version string in here be for 17a or an older release?  I'm just going by the name of this test case and I'm not sure if I'm interpreting this correctly.

src/test/java/com/mathworks/ci/MatlabBuilderTest.java
```diff
@@ -279,4 +282,134 @@ public void verifyCustomCommandInvokedForBatchMode() throws Exception {
...
+        RunTestsAutomaticallyOption runOption = new RunTestsAutomaticallyOption();
+        runOption.setTaCoberturaChkBx(true);
+        runOption.setTaJunitChkBx(true);
+        runOption.setTatapChkBx(true);
+        this.matlabBuilder.setTestRunTypeList(runOption);
+        project.getBuildersList().add(this.matlabBuilder);
+        FreeStyleBuild build = project.scheduleBuild2(0).get();
+        jenkins.assertLogContains("-batch", build);
+        jenkins.assertLogContains("true,true,true", build);
```
jwilkick-mw: Is there any chance this test could pick up the wrong output?  Is there some way to make it clearer that these boolean values came from the checkboxes?

src/test/java/com/mathworks/ci/MatlabBuilderTest.java
```diff
@@ -279,4 +282,134 @@ public void verifyCustomCommandInvokedForBatchMode() throws Exception {
...
+    
+    /*
+     * Test to verify default value of getStringByName() when Automatic test mode. 
+     */
+
+    @Test
+    public void verifyDefaultValueOfgetStringByName() throws Exception {
+        this.matlabBuilder.setLocalMatlab(getMatlabroot("R2018b"));
+        RunTestsAutomaticallyOption runOption = new RunTestsAutomaticallyOption();    
+        Assert.assertNull(runOption.getStringByName("fakeChkBox"));
```
jwilkick-mw: It seems like this is testing the behavior of a different class. Is there a separate test for the RunTestsAutomaticallyOption that could do this?

src/test/java/com/mathworks/ci/MatlabBuilderTest.java
```diff
@@ -279,4 +282,134 @@ public void verifyCustomCommandInvokedForBatchMode() throws Exception {
...
+
+    @Test
+    public void verifyMatlabVersionOlderThanR17a() throws Exception {
+        this.matlabBuilder.setLocalMatlab(getMatlabroot("R2018b"));
+        RunTestsAutomaticallyOption runOption = new RunTestsAutomaticallyOption();
+        runOption.setTaCoberturaChkBx(true);
+        runOption.setTaJunitChkBx(true);
+        runOption.setTatapChkBx(true);
+        this.matlabBuilder.setTestRunTypeList(runOption);
+        project.getBuildersList().add(this.matlabBuilder);
```
jwilkick-mw: Noticed that some of the code from this point to the end of the method is repeated in another test.  Could you pull this out into a different function?

src/test/java/com/mathworks/ci/MatlabBuilderTest.java
```diff
@@ -279,4 +282,134 @@ public void verifyCustomCommandInvokedForBatchMode() throws Exception {
...
+        WebAssert.assertTextPresent(page, TestMessage.getValue("Builder.matlab.root.empty.error"));
+    }
+    
+    /*
+     * Test To verify UI does not throw any error when valid MATLAB root entered
+     * 
+     */
+
+    @Test
+    public void verifyInvalidValidMatlabRoot() throws Exception {
```
jwilkick-mw: Given the comment, could you name this:  verifyValidMatlabRootDoesntDisplayError?

Or, looking at the actual test code, it seems like it **is** checking that an invalid root shows an error. I'm a bit confused by this one.

src/test/java/com/mathworks/ci/MatlabBuilderTest.java
```diff
@@ -279,4 +282,134 @@ public void verifyCustomCommandInvokedForBatchMode() throws Exception {
...
+        WebAssert.assertTextPresent(page, TestMessage.getValue("Builder.matlab.cobertura.support.warning"));
+    }
+    
+    /*
+     * Test To verify UI displays Cobertura Error message when invalid MATLAB root entered.
+     * 
+     */
+
+    @Test
+    public void verifyCoberturaError() throws Exception {
```
jwilkick-mw: Should this be testing that Builder.matlab.cobertura.support.warning is there?  Otherwise it seems like the invalid matlab root test might cover this already.

src/test/java/com/mathworks/ci/MatlabBuilderTest.java
nbhoski: Thanks for pointing out this has to be older version. fixed it by using R2016b version. it required some change in the getMatlabroot() method as well.

src/test/java/com/mathworks/ci/MatlabBuilderTest.java
nbhoski: these have chance of taking default values if someone changes the .jelly file with corresponding web element names , but forgets to add it in action class. i don't see direct way to map these to confirm if its coming from UI other than writing some Web automation (functional tests) . I got your point though.

src/test/java/com/mathworks/ci/MatlabBuilderTest.java
nbhoski: No , we do not have separate test for this class as this is part of MatlabBuilder class by design. to support multi dropdown.options . this test actually is targeting more of line coverage / line coverage.

src/test/java/com/mathworks/ci/MatlabBuilderTest.java
nbhoski: Done.

src/test/java/com/mathworks/ci/MatlabBuilderTest.java
nbhoski: Good catch , its copy paste error, This test case checks if error is displayed if invalid MATLAB root is entered. i have changed the name to verifyInvalidMatlabRootDisplaysError and verifyValidMatlabRootDoesntDisplayError

src/test/java/com/mathworks/ci/MatlabBuilderTest.java
nbhoski: Correct , this Error is already verified however both these scenario uses MatlabReleaseInfo to check the MATLAb version and both of them they throw version not found exception. hence in the exception case our code coverage was not hitting the exception block. to cover the same this test is added.

src/test/java/com/mathworks/ci/MatlabBuilderTest.java
```diff
@@ -87,9 +88,17 @@ public void testTearDown() {
 
     private String getMatlabroot(String version) throws URISyntaxException {
         ClassLoader classLoader = MatlabBuilderTest.class.getClassLoader();
-        String matlabRoot = new File(
-                classLoader.getResource("versioninfo/" + version + "/VersionInfo.xml").toURI())
-                        .getAbsolutePath().replace(FileSeperator + "VersionInfo.xml", "");
+        String matlabRoot;
```
jwilkick-mw: This bit of code is kind of hard to follow.  Could you assign the result of `classLoader.getResource(...)` to a variable and then use that in the if condition?

```java
String matlabRoot;
URL versionInfo = classLoader.getResource("versioninfo/"+version+"/VersionInfo.xml");

if(versionInfo!=null){
    matlabRoot = new File(versionInfo.toURI()).getAbsolutePath();
}else{
    matlabRoot = new File(classLoader.getResource("versioninfo/R2017a/VersionInfo.xml").toURI()).getAbsolutePath().replace("R2017a",version);
}

return matlabRoot.replace(FileSeparator+"VersionInfo.xml","");
```
Not everyone is a fan of Optionals, but I could also see the code like this:

```java
private String getMatlabroot(String version) throws URISyntaxException {
     String defaultVersionInfo = "versioninfo/R2017a/VersionInfo.xml";
     String userVersionInfo = "versioninfo/"+version+"/VersionInfo.xml";

      Optional<URL> matlabRootURL = Optional.ofNullable(getResource(userVersionInfo)).orElseGet(() -> getResource(defaultVersionInfo));
      
      File matlabRoot = new File(matlabRootURL.get().toURI());

      return matlabRoot.getAbsolutePath().replace(FileSeperator+"VersionInfo.xml","").replace("R2017a",version);      
}

private URL getResource(String resource){
    return MatlabBuilderTest.class.getClassLoader().getResource(resource);   
}
```

src/test/java/com/mathworks/ci/MatlabBuilderTest.java
```diff
@@ -412,4 +409,19 @@ public void verifyCoberturaError() throws Exception {
         String filteredPageText = pageText.replaceFirst(TestMessage.getValue("Builder.invalid.matlab.root.error"), "");
         Assert.assertTrue(filteredPageText.contains(TestMessage.getValue("Builder.invalid.matlab.root.error")));
     }
+    
+    /*
+     * Private helper methods for tests
+     */
+    
+    private FreeStyleBuild getBuildforRunTestAutomatically() throws InterruptedException, ExecutionException {
```
jwilkick-mw: The helper method is a good addition.

src/test/java/com/mathworks/ci/MatlabBuilderTest.java
nbhoski: Thanks @jwilkick-mw  . it was a good suggestion for using Optional <T> class

PR #24, opened 2019-06-06
Custom command is run after escaping imbalanced double quotes

nbhoski:
custom commands are run using eval() function 
escaping all double quotes to get appropriate command

### DIFF
--- a/src/main/java/com/mathworks/ci/MatlabBuilder.java
+++ b/src/main/java/com/mathworks/ci/MatlabBuilder.java
@@ -471,8 +471,8 @@ public class MatlabBuilder extends Builder implements SimpleBuildStep {
                     + getTestRunTypeList().getBooleanByName("taCoberturaChkBx")
                     + ")),catch e,disp(getReport(e,'extended')),exit(1),end";
         } else {
-            runCommand = "try," + this.getTestRunTypeList().getStringByName("customMatlabCommand")
-                    + ",catch e,disp(getReport(e,'extended')),exit(1),end,exit";
+            runCommand = "try,eval(\"" + this.getTestRunTypeList().getStringByName("customMatlabCommand").replaceAll("\"","\"\"")
+                    + "\"),catch e,disp(getReport(e,'extended')),exit(1),end,exit";
         }
 
         final String[] runnerSwitch = {"-r", runCommand};
--- a/src/test/java/com/mathworks/ci/MatlabBuilderTest.java
+++ b/src/test/java/com/mathworks/ci/MatlabBuilderTest.java
@@ -267,7 +267,7 @@ public class MatlabBuilderTest {
         project.getBuildersList().add(this.matlabBuilder);
         FreeStyleBuild build = project.scheduleBuild2(0).get();
         jenkins.assertLogContains("-r", build);
-        jenkins.assertLogContains("try,runtests", build);
+        jenkins.assertLogContains("try,eval", build);
     }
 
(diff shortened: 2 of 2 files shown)


src/test/java/com/mathworks/ci/MatlabBuilderTest.java
```diff
@@ -267,7 +267,7 @@ public void verifyCustomCommandInvoked() throws Exception {
         project.getBuildersList().add(this.matlabBuilder);
         FreeStyleBuild build = project.scheduleBuild2(0).get();
         jenkins.assertLogContains("-r", build);
-        jenkins.assertLogContains("try,runtests", build);
+        jenkins.assertLogContains("try,eval", build);
```
jwilkick-mw: I think it might be useful to try a few particular test cases for the custom string, like:   ") disp('hello') eval("

src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -471,8 +471,8 @@ private int execMatlabCommand(Run<?, ?> build, FilePath workspace, Launcher laun
                     + getTestRunTypeList().getBooleanByName("taCoberturaChkBx")
                     + ")),catch e,disp(getReport(e,'extended')),exit(1),end";
         } else {
-            runCommand = "try," + this.getTestRunTypeList().getStringByName("customMatlabCommand")
-                    + ",catch e,disp(getReport(e,'extended')),exit(1),end,exit";
+            runCommand = "try,eval(\"" + this.getTestRunTypeList().getStringByName("customMatlabCommand").replaceAll("\"","\"\"")
```
acampbel: I don't think we have resolved Joe's concerns about security on this yet. I think it is great to see that this nominally works and may be an option, but I am not ready to say ship it until we resolve that conversation

nbhoski: I also would like draw your attention to the MATLAB's its own ability to [run any shell command using !](https://www.mathworks.com/help/matlab/matlab_external/run-external-commands-scripts-and-programs.html) . If we allow custom command and user chooses to execute the shell command using this MATLAB feature we will still have the same concerns.

nbhoski: @jwilkick-mw . Would it be a good idea if we could try to reproduce the issues on jenkins with above implementation . I shall share .hpi file and a Jenkins instance with this implementation.

jwilkick-mw: I played around with it a little bit.  With the input string ") disp('hello') eval(", I got:

ans=
  " ) disp('hello') eval("

which seems pretty reasonable.  I played around with various combinations of using backslashes and even unicode escapes and the closest I came to breaking it was the input:  \")disp('hello')eval("\"   but MATLAB just caught that as an error.  I think this fix might be enough to mitigate breaking out; the whole command is single-quoted as well, so the shell shouldn't try to interpret it.  

The shell MATLAB command does still seem to be a possible concern. Could we have the plugin look for the ! in the custom MATLAB command and refuse to run the command if it finds it?  How does MATLAB Online handle the ! operator?

nbhoski:
@jwilkick-mw  @acampbel Shall i merge this PR based on our conversation.

acampbel:
> @jwilkick-mw @acampbel Shall i merge this PR based on our conversation.

Ship It!

Issue #23, opened 2019-05-28
the plug-in uses a R2016a onwards function


versionbayjc:
From the documentation I understood this plug-in would work in R2013a and newer, however when I use the automatic test mode script manually in R2013a, it fails with:
```
>> runMatlabTests(true,true,false)
Undefined function 'testsuite' for input arguments of type 'char'.

Error in runMatlabTests (line 15)
suite = testsuite(pwd,'IncludeSubfolders',true);
```

And indeed if I look at the doc, "testsuite" only exists from R2016a onwards. Maybe the automatic test mode is not designed to work with releases of R2015b and older?

It seems the plug-in does recognize the MATLAB version, because it does not allow selecting Cobertura code coverage for older than R2017b.

acampbel:
The automatic mode should indeed apply to releases older than R2016a.

nbhoski:
Thanks @versionbayjc for bringing this to our notice. Working on it

nbhoski:
fix is merged with number #25

PR #25, opened 2019-07-15
testsuite support for MATLAB version less than R2016a

nbhoski:
added support for creating test suites for MATLAB versions less than R2016a

### DIFF
--- a/src/main/resources/com/mathworks/ci/MatlabBuilder/runMatlabTests.m
+++ b/src/main/resources/com/mathworks/ci/MatlabBuilder/runMatlabTests.m
@@ -3,6 +3,7 @@
 function failed = runMatlabTests(produceJUnit, produceTAP, produceCobertura)
 
 BASE_VERSION_MATLABUNIT_SUPPORT = '8.1';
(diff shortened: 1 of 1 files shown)


nbhoski:
fix for #23

src/main/resources/com/mathworks/ci/MatlabBuilder/runMatlabTests.m
```diff
@@ -12,10 +13,14 @@
 import('matlab.unittest.TestSuite');
 
 %Create test suite for tests folder
-suite = testsuite(pwd,'IncludeSubfolders',true);
+if verLessThan('matlab',BASE_VERSION_TESTSUITE_SUPPORT)
```
mw-akumar: Would it make it easier to read if this logic was refactored out to a separate local function. 

the code in runMATLABTests would simply become

suite = gettestsuite(); 

And the special case logic related to selection of the correct testsuite as per release will be in a its own function.

nbhoski: Done.

PR #31, opened 2019-08-26
Readme revision

mw-hrastega:
Hi, I have revised the readme text. Please review and share your comments.
-Houman


README.md
```diff
@@ -1,113 +1,105 @@
 # Jenkins MATLAB Plugin
 
-## Description
-The Jenkins plugin for MATLAB&reg; enables you to easily run your MATLAB tests and generate test artifacts in formats such as JUnit, TAP, and Cobertura code coverage reports. 
-## Build Step Configuration
-To invoke this plugin, select "Run MATLAB Tests" from the Add build step list.
+The Jenkins plugin for MATLAB&reg; enables you to easily run your MATLAB tests and generate test artifacts in formats such as *JUnit*, *TAP*, and *Cobertura* code coverage reports. You can run tests in either *automatic* or *custom* mode.
```
nbhoski: Its a good idea to inform users about the run modes. However, the user may assume that artifacts generation is possible in both modes. where we, actually support artifact generation only for the Automatic option. could we rephrase it something like?

The Jenkins plugin for MATLAB enables you to easily run your MATLAB tests and generate test artifacts in formats such as JUnit, TAP, and Cobertura code coverage reports using automatic mode. You can also run tests in custom mode.

nbhoski: We can generate artifacts by customizing but we automatically do no generate artifacts for Custom options. User should if he/she wants one and write script accordingly.

mw-hrastega: I removed the sentence "You can run tests in either *automatic* or *custom* mode" and incorporated its information into the first sentence of the section "Option 1: Freestyle project with automatic test mode". Let me know if it is clearer this time.

PR #34, opened 2019-09-27
Scratch file copy operation restricted to Automatic mode only.

nbhoski:
Scratch file generation will happen only for Automatic mode.

### DIFF
--- a/src/main/java/com/mathworks/ci/MatlabBuilder.java
+++ b/src/main/java/com/mathworks/ci/MatlabBuilder.java
@@ -59,6 +59,7 @@ public class MatlabBuilder extends Builder implements SimpleBuildStep {
             "Builder.matlab.runner.target.file.name";
     private static final String MATLAB_RUNNER_RESOURCE =
             "com/mathworks/ci/MatlabBuilder/runMatlabTests.m";
+    private static final String AUTOMATIC_OPTION = "RunTestsAutomaticallyOption";
 
 
     @DataBoundConstructor
@@ -408,11 +409,10 @@ public class MatlabBuilder extends Builder implements SimpleBuildStep {
             TaskListener listener, boolean isLinuxLauncher)
             throws IOException, InterruptedException {
         setEnv(build.getEnvironment(listener));
-        final String testRunMode = this.getTestRunTypeList().getDescriptor().getDisplayName();
+        final String testRunMode = this.getTestRunTypeList().getDescriptor().getId();
         
         // Copy MATLAB scratch file into the workspace only if Automatic option is selected.
-        if (!testRunMode.equalsIgnoreCase(
-                Message.getValue("builder.matlab.customcommandoption.display.name"))) {
+        if (testRunMode.contains(AUTOMATIC_OPTION)) {
             copyMatlabScratchFileInWorkspace(MATLAB_RUNNER_RESOURCE, MATLAB_RUNNER_TARGET_FILE,
                     workspace, getClass().getClassLoader());
         }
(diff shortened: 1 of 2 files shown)


nbhoski:
All checks have passed 
https://dev.azure.com/iat-ci/jenkins-matlab-plugin/_build/results?buildId=149

src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -428,12 +430,16 @@ private synchronized int execMatlabCommand(Run<?, ?> build, FilePath workspace,
         return matlabLauncher.join();
     }
 
-    public List<String> constructMatlabCommandWithBatch() {
+    public List<String> constructMatlabCommandWithBatch() throws IOException, InterruptedException {
```
mw-akumar: It seems odd that a function called constructMatlabCommandWithBatch should throw an IOException. This looks like the exception specification of this method is determined by the implementation details of this method. 

Similarly, methods whose job is to construct commands (based on -r or -batch) is now tasked with the additional responsibility of doing copy of the scratch file. That seems out of place as well.

src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -452,7 +458,7 @@ private synchronized int execMatlabCommand(Run<?, ?> build, FilePath workspace,
         return matlabDefaultArgs;
     }
 
-    public List<String> constructDefaultMatlabCommand(boolean isLinuxLauncher) {
+    public List<String> constructDefaultMatlabCommand(boolean isLinuxLauncher) throws IOException, InterruptedException {
```
mw-akumar: same command as earlier.

src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -479,11 +485,15 @@ private synchronized int execMatlabCommand(Run<?, ?> build, FilePath workspace,
...
     }
 
-    private String[] getRunnerSwitch() {
+    private String[] getRunnerSwitch() throws IOException, InterruptedException {
         final String runCommand;
         final String testRunMode = this.getTestRunTypeList().getDescriptor().getDisplayName();
         if (!testRunMode.equalsIgnoreCase(
                 Message.getValue("builder.matlab.customcommandoption.display.name"))) {
+            
+            //Copy MATLAB scratch file into the workspace only if Automatic option is selected.
```
mw-akumar: A method called getRunnerSwitch (whose responsibility is to provide the switches used by runner) is doing a copy operation for setting up the artifacts required to run the tests. This seems odd.

src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -408,9 +412,7 @@ private synchronized int execMatlabCommand(Run<?, ?> build, FilePath workspace,
             TaskListener listener, boolean isLinuxLauncher)
             throws IOException, InterruptedException {
         setEnv(build.getEnvironment(listener));
-        //Copy MATLAB scratch file into the workspace
-        copyMatlabScratchFileInWorkspace(MATLAB_RUNNER_RESOURCE, MATLAB_RUNNER_TARGET_FILE,
```
mw-akumar: Ideally we could just find a way to conditionally do this copy based on the test run mode. This will avoid the need to inject the knowledge of scratch file down in low level functions who are trying to determine the command based on -r and -batch.

src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -408,9 +408,14 @@ private synchronized int execMatlabCommand(Run<?, ?> build, FilePath workspace,
...
-                workspace, getClass().getClassLoader());
+        final String testRunMode = this.getTestRunTypeList().getDescriptor().getDisplayName();
```
acampbel: IS there a way we don't have to look at the display name for this? This seems like it is coupling the view with what should be a model concern.

src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -408,9 +408,14 @@ private synchronized int execMatlabCommand(Run<?, ?> build, FilePath workspace,
...
+        
+        // Copy MATLAB scratch file into the workspace only if Automatic option is selected.
+        if (!testRunMode.equalsIgnoreCase(
```
acampbel: Curious why we dont do this operation only in automatic mode vs when it is not custom mode. AUtomatic seems more direct and makes less assumptions.

acampbel: In general it seems like this code could be refactored to avoid the code branches based on these two modes. Somethign to consider as we break out the build into two steps.

PR #43, opened 2019-11-14
Added changes to support running MATLAB on remote agents.

nbhoski:
Added changes to support remote nodes on Jenkins plugin. with these changes now user can run MATLAB on Jenkins agents.

### DIFF
--- a/src/main/java/com/mathworks/ci/MatlabBuilder.java
+++ b/src/main/java/com/mathworks/ci/MatlabBuilder.java
@@ -60,7 +60,7 @@ public class MatlabBuilder extends Builder implements SimpleBuildStep {
     private static final String MATLAB_RUNNER_RESOURCE =
             "com/mathworks/ci/MatlabBuilder/runMatlabTests.m";
     private static final String AUTOMATIC_OPTION = "RunTestsAutomaticallyOption";
-    private String fileSeparator;
+    private String nodeSpecificfileSeparator;
 
 
     @DataBoundConstructor
@@ -103,13 +103,6 @@ public class MatlabBuilder extends Builder implements SimpleBuildStep {
         this.env = env;
     }
     
-    private void setFileSeparator(String fileSeparator) {
-        this.fileSeparator = fileSeparator;
-    }
-    
-    private String getFileSeparator() {
-        return this.fileSeparator;
-    }
 
     @Extension
     public static class MatlabDescriptor extends BuildStepDescriptor<Builder> {
@@ -403,8 +396,7 @@ public class MatlabBuilder extends Builder implements SimpleBuildStep {
             @Nonnull Launcher launcher, @Nonnull TaskListener listener)
             throws InterruptedException, IOException {        
         final boolean isLinuxLauncher = launcher.isUnix();
-        String fileSeparator = getNodeSpecificFileSeperator(launcher);
-        this.setFileSeparator(fileSeparator);
+        nodeSpecificfileSeparator = getNodeSpecificFileSeperator(launcher);
         
         // Invoke MATLAB command and transfer output to standard
         // Output Console
@@ -421,11 +413,11 @@ public class MatlabBuilder extends Builder implements SimpleBuildStep {
             throws IOException, InterruptedException {
         setEnv(build.getEnvironment(listener));
         final String testRunMode = this.getTestRunTypeList().getDescriptor().getId();
+        FilePath targetWorkspace = new FilePath(launcher.getChannel(), workspace.getRemote());
         
         // Copy MATLAB scratch file into the workspace only if Automatic option is selected.
         if (testRunMode.contains(AUTOMATIC_OPTION)) {
-            copyMatlabScratchFileInWorkspace(MATLAB_RUNNER_RESOURCE, MATLAB_RUNNER_TARGET_FILE,
-                    workspace, getClass().getClassLoader(), launcher);
+            copyMatlabScratchFileInWorkspace(MATLAB_RUNNER_RESOURCE, MATLAB_RUNNER_TARGET_FILE, targetWorkspace);
         }
         ProcStarter matlabLauncher;
         try {
@@ -462,7 +454,7 @@ public class MatlabBuilder extends Builder implements SimpleBuildStep {
         }
 
         matlabDefaultArgs =
-                Arrays.asList(getLocalMatlab() + this.getFileSeparator() + "bin" + this.getFileSeparator() + "matlab",
+                Arrays.asList(getLocalMatlab() + nodeSpecificfileSeparator + "bin" + nodeSpecificfileSeparator + "matlab",
                         "-batch", runCommand);
         
         return matlabDefaultArgs;
@@ -485,7 +477,7 @@ public class MatlabBuilder extends Builder implements SimpleBuildStep {
 
     private String[] getPreRunnerSwitches() {
         String[] preRunnerSwitches =
-                {getLocalMatlab() + this.getFileSeparator() + "bin" + this.getFileSeparator() + "matlab", "-nosplash",
+                {getLocalMatlab() + nodeSpecificfileSeparator + "bin" + nodeSpecificfileSeparator + "matlab", "-nosplash",
                         "-nodesktop", "-noAppIcon"};
         return preRunnerSwitches;
     }
@@ -516,9 +508,9 @@ public class MatlabBuilder extends Builder implements SimpleBuildStep {
     }
     
     private void copyMatlabScratchFileInWorkspace(String matlabRunnerResourcePath,
-            String matlabRunnerTarget, FilePath workspace, ClassLoader classLoader, Launcher launcher)
+            String matlabRunnerTarget, FilePath targetWorkspace)
             throws IOException, InterruptedException {
-        FilePath targetWorkspace = new FilePath(launcher.getChannel(), workspace.getRemote());
+        final ClassLoader classLoader = getClass().getClassLoader();
         FilePath targetFile =
                 new FilePath(targetWorkspace, Message.getValue(matlabRunnerTarget));
         InputStream in = classLoader.getResourceAsStream(matlabRunnerResourcePath);
--- a/src/main/java/com/mathworks/ci/MatlabBuilder.java
+++ b/src/main/java/com/mathworks/ci/MatlabBuilder.java
@@ -60,6 +60,7 @@ public class MatlabBuilder extends Builder implements SimpleBuildStep {
     private static final String MATLAB_RUNNER_RESOURCE =
             "com/mathworks/ci/MatlabBuilder/runMatlabTests.m";
     private static final String AUTOMATIC_OPTION = "RunTestsAutomaticallyOption";
+    private String fileSeparator;
 
 
     @DataBoundConstructor
@@ -100,6 +101,14 @@ public class MatlabBuilder extends Builder implements SimpleBuildStep {
     }
     private void setEnv(EnvVars env) {
         this.env = env;
+    }
+    
+    private void setFileSeparator(String fileSeparator) {
+        this.fileSeparator = fileSeparator;
+    }
+    
+    private String getFileSeparator() {
+        return this.fileSeparator;
     }
 
     @Extension
@@ -197,7 +206,7 @@ public class MatlabBuilder extends Builder implements SimpleBuildStep {
                     }
                 } catch (MatlabVersionNotFoundException e) {
                     return FormValidation
-                            .error(Message.getValue("Builder.invalid.matlab.root.error"));
+                            .warning(Message.getValue("Builder.invalid.matlab.root.warning"));
                 }
             }
             return FormValidation.ok();
@@ -266,7 +275,7 @@ public class MatlabBuilder extends Builder implements SimpleBuildStep {
                                 .warning(Message.getValue("Builder.matlab.cobertura.support.warning"));
                     }
                 } catch (MatlabVersionNotFoundException e) {
-                    return FormValidation.error(Message.getValue("Builder.invalid.matlab.root.error"));
+                    return FormValidation.warning(Message.getValue("Builder.invalid.matlab.root.warning"));
                 }
             }
             
@@ -394,6 +403,8 @@ public class MatlabBuilder extends Builder implements SimpleBuildStep {
             @Nonnull Launcher launcher, @Nonnull TaskListener listener)
             throws InterruptedException, IOException {        
         final boolean isLinuxLauncher = launcher.isUnix();
+        String fileSeparator = getNodeSpecificFileSeperator(launcher);
+        this.setFileSeparator(fileSeparator);
         
         // Invoke MATLAB command and transfer output to standard
         // Output Console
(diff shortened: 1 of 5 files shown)


nbhoski:
Azzure job [# 20191114.1](https://dev.azure.com/iat-ci/jenkins-matlab-plugin/_build/results?buildId=160) passes for this PR

src/main/resources/config.properties
```diff
@@ -4,7 +4,7 @@
 Builder.display.name = Run MATLAB Tests
 Builder.matlab.runner.target.file.name = runMatlabTests.m
 Builder.matlab.cobertura.support.warning = To generate a Cobertura report, use MATLAB R2017b or a newer version.
-Builder.invalid.matlab.root.error = Unable to launch MATLAB from the specified location. Verify the MATLAB root folder path.
+Builder.invalid.matlab.root.warning = Unable to find MATLAB from the specified location on this system(but perhaps it exists on some agents)
```
mcafaro: This wording is awkward. Have Houman take a look. My attempt:

"Unable to determine if MATLAB exists at the specified location."

src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -101,6 +102,14 @@ private String getCustomMatlabCommand() {
...
         this.env = env;
     }
+    
+    private void setFileSeparator(String fileSeparator) {
+        this.fileSeparator = fileSeparator;
+    }
+    
+    private String getFileSeparator() {
+        return this.fileSeparator;
+    }
```
mcafaro: Getters/setters for private variables seem unnecessary. Why not just access the private field directly?

src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -450,7 +462,7 @@ private synchronized int execMatlabCommand(Run<?, ?> build, FilePath workspace,
         }
 
         matlabDefaultArgs =
-                Arrays.asList(getLocalMatlab() + File.separator + "bin" + File.separator + "matlab",
+                Arrays.asList(getLocalMatlab() + this.getFileSeparator() + "bin" + this.getFileSeparator() + "matlab",
```
mcafaro: You can remove the `this.`

src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -473,7 +485,7 @@ private synchronized int execMatlabCommand(Run<?, ?> build, FilePath workspace,
 
     private String[] getPreRunnerSwitches() {
         String[] preRunnerSwitches =
-                {getLocalMatlab() + File.separator + "bin" + File.separator + "matlab", "-nosplash",
+                {getLocalMatlab() + this.getFileSeparator() + "bin" + this.getFileSeparator() + "matlab", "-nosplash",
```
mcafaro: You can remove the `this.`

src/main/java/com/mathworks/ci/MatlabReleaseInfo.java
```diff
@@ -112,4 +99,56 @@ else if(!new File(this.matlabRoot).exists()){
...
+        
+    private static final class RemoteFileOperation implements FileCallable<Map<String, String>> {
+
+        private static final long serialVersionUID = 1L;
+
+        private Map<String, String> versionInfoCache = new HashMap<String, String>();
+
+        @Override
+        public void checkRoles(RoleChecker checker) throws SecurityException {
+
```
mcafaro: If this is intended to be empty, consider adding a comment saying so. Maybe: `// no checks to perform`

src/main/resources/config.properties
nbhoski: Added @mw-hrastega

src/main/java/com/mathworks/ci/MatlabBuilder.java
nbhoski: Done

src/main/java/com/mathworks/ci/MatlabBuilder.java
nbhoski: Done

src/test/resources/testconfig.properties
```diff
@@ -4,5 +4,5 @@
 Verify.matlab.invokes.positive = MATLAB is invoking positive tests
```
mw-hrastega: What do you mean by positive test?

src/test/resources/testconfig.properties
```diff
@@ -4,5 +4,5 @@
...
 Verify.build.ignore.test.failure = Build Ignored test failure
```
mw-hrastega: Why is Ignored capitalized? Do you mean "Build ignored the test failure"?

src/test/resources/testconfig.properties
```diff
@@ -4,5 +4,5 @@
...
 Builder.matlab.cobertura.support.warning = To generate a Cobertura report, use MATLAB R2017b or a newer version.
```
mw-hrastega: newer version --> newer release

src/test/resources/testconfig.properties
```diff
@@ -4,5 +4,5 @@
...
-Builder.invalid.matlab.root.error = Unable to launch MATLAB from the specified location. Verify the MATLAB root folder path.
+Builder.invalid.matlab.root.warning = Unable to find MATLAB from the specified location on this system(but perhaps it exists on some agents)
```
mw-hrastega: Unable to open MATLAB from the specified folder. 

Why do we want to provide this information? "(but perhaps it exists on some agents)"

src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -410,15 +413,16 @@ private synchronized int execMatlabCommand(Run<?, ?> build, FilePath workspace,
             throws IOException, InterruptedException {
```
mw-akumar: Let's try to reduce the level of detail in this top level function. This function should read like the business logic of our plugin implementation

1. Determine the platform on which we will run (Input: perform method)
2. Determine the channel
3. Determine the MATLAB root on the remote machine (Input: Output of 2, class property)
4. Determine the MATLAB version (Inputs : output of 3)
5. Construct the MATLAB command (Inputs: Output of 4 and output of 1)
6. Determine the node workspace
7. Invoke the MATLAB command (Inputs: Output of 5 and 6)

src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -410,15 +413,16 @@ private synchronized int execMatlabCommand(Run<?, ?> build, FilePath workspace,
...
         setEnv(build.getEnvironment(listener));
```
mw-akumar: this goes closer to the creation of the ProcLauncher

src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -410,15 +413,16 @@ private synchronized int execMatlabCommand(Run<?, ?> build, FilePath workspace,
...
         final String testRunMode = this.getTestRunTypeList().getDescriptor().getId();
+        FilePath targetWorkspace = new FilePath(launcher.getChannel(), workspace.getRemote());
```
mw-akumar: moves closer to the code where we launch and execute

src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -410,15 +413,16 @@ private synchronized int execMatlabCommand(Run<?, ?> build, FilePath workspace,
...
         if (testRunMode.contains(AUTOMATIC_OPTION)) {
-            copyMatlabScratchFileInWorkspace(MATLAB_RUNNER_RESOURCE, MATLAB_RUNNER_TARGET_FILE,
-                    workspace, getClass().getClassLoader());
+            copyMatlabScratchFileInWorkspace(MATLAB_RUNNER_RESOURCE, MATLAB_RUNNER_TARGET_FILE, targetWorkspace);
         }
         ProcStarter matlabLauncher;
         try {
-            MatlabReleaseInfo rel = new MatlabReleaseInfo(getLocalMatlab());
+            FilePath nodeSpecificMatlabRoot = new FilePath(launcher.getChannel(),getLocalMatlab());
+            MatlabReleaseInfo rel = new MatlabReleaseInfo(nodeSpecificMatlabRoot);
```
mw-akumar: since the only exception being caught is one coming from MatlabReleaseInfo, only this code needs to be in the try.

src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -410,15 +413,16 @@ private synchronized int execMatlabCommand(Run<?, ?> build, FilePath workspace,
...
             matlabLauncher = launcher.launch().pwd(workspace).envs(this.env);
             if (rel.verLessThan(BASE_MATLAB_VERSION_BATCH_SUPPORT)) {
```
mw-akumar: matlab command construction should be its own methods and should be able to work with a couple of inputs.

src/test/resources/testconfig.properties
nbhoski: Yes, Build ignored the test failure is correct. changing the capitalization

src/test/resources/testconfig.properties
nbhoski: Done.

src/test/resources/testconfig.properties
nbhoski: This is due to remote slave support feature where entered MATLAB root may not be present on the master node however its possible that user must have entered the MATLAB root present on remote agent.

src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -101,6 +102,14 @@ private String getCustomMatlabCommand() {
     private void setEnv(EnvVars env) {
         this.env = env;
     }
+    
+    private void setFileSeparator(String fileSeparator) {
```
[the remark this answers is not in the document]

src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -504,13 +516,22 @@ private synchronized int execMatlabCommand(Run<?, ?> build, FilePath workspace,
     }
     
     private void copyMatlabScratchFileInWorkspace(String matlabRunnerResourcePath,
```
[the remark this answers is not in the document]
nbhoski: Done. creating an appropriate target workspace.

src/main/java/com/mathworks/ci/MatlabBuilder.java
nbhoski: Done. refactored the code as close as possible to read it in the suggested way.

src/main/java/com/mathworks/ci/MatlabBuilder.java
nbhoski: in order to keep the execMatlab command parameters minimalistic, I have moved this up in the perform method.

src/main/java/com/mathworks/ci/MatlabBuilder.java
nbhoski: The scope of Exception is widened now with the copy operation. and the intention is to catch any exception and exit with an appropriate exit status. changes the exception type accordingly .

src/main/java/com/mathworks/ci/MatlabBuilder.java
nbhoski: Refactored to read in procedural way that is suggested

src/test/resources/testconfig.properties
nbhoski: These are tests which will not fail and are straight forward scenarios.

src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -257,7 +260,8 @@ public FormValidation doCheckTaCoberturaChkBx(@QueryParameter boolean taCobertur
         }
 
         Function<String, FormValidation> chkCoberturaSupport = (String matlabRoot) -> {
-            rel = new MatlabReleaseInfo(matlabRoot);
+            FilePath matlabRootPath = new FilePath(new File(matlabRoot));
```
mw-akumar: If matlabroot is always consumed by first creating a FilePath object, then it make sense to keep the FilePath object at the same scope as the matlabroot (or even replace it if needed).

src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -393,40 +397,44 @@ public String getStringByName(String memberName) {
...
-        
-        // Copy MATLAB scratch file into the workspace only if Automatic option is selected.
-        if (testRunMode.contains(AUTOMATIC_OPTION)) {
-            copyMatlabScratchFileInWorkspace(MATLAB_RUNNER_RESOURCE, MATLAB_RUNNER_TARGET_FILE,
-                    workspace, getClass().getClassLoader());
-        }
         ProcStarter matlabLauncher;
         try {
-            MatlabReleaseInfo rel = new MatlabReleaseInfo(getLocalMatlab());
+            FilePath nodeSpecificMatlabRoot = new FilePath(launcher.getChannel(),getLocalMatlab());
```
mw-akumar: similar comment as earlier. If the only way the local matlab is usable is by wrapping in a FilePath object, then just have it give a FilePath object.

src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -393,40 +397,44 @@ public String getStringByName(String memberName) {
...
-                matlabLauncher = matlabLauncher.cmds(constructDefaultMatlabCommand(isLinuxLauncher)).stderr(outStream);
+                matlabLauncher = matlabLauncher.cmds(constructDefaultMatlabCommand(launcher.isUnix())).stderr(outStream);
             } else {
                 matlabLauncher = matlabLauncher.cmds(constructMatlabCommandWithBatch()).stdout(listener);
             }
-        } catch (MatlabVersionNotFoundException e) {
+            
+            //Check the test run mode option selected by user and identify the target workspace to copy the scratch file.
+            final String testRunMode = this.getTestRunTypeList().getDescriptor().getId();
+            FilePath targetWorkspace = new FilePath(launcher.getChannel(), workspace.getRemote());
```
mw-akumar: construction of targetworkspace should be done only in the scope that is needed

src/main/java/com/mathworks/ci/MatlabBuilder.java
nbhoski: where mostly getLocalMatlab() is used wrapped in FilePath(). its not always true. getLocalMatlab()  is wrapped in FilePath only when we want to now the versions of the MATLAB. otherside in some cases we just would just require getLocalMAtlab() out put with different way of FilPath construction. for eg this case needs FilePath construction with Launcher  and file as string . everywhere else where we just need expanded Matlabroot, we will just use getLocalMatlab() as is.

src/main/java/com/mathworks/ci/MatlabBuilder.java
nbhoski: captured this as task. may have to call it using reflections if i change the scope of this outside of this static class. captured it. i will discuss the possible solutions during commits of the captured task for the same.

nbhoski:
Build on azzure [#20191202.1](https://dev.azure.com/iat-ci/jenkins-matlab-plugin/_build/results?buildId=167) passed. hence merging it in master.

PR #62, opened 2020-01-06
Version check for noAppIcon startup option.

nbhoski:
Adding noAppIcon startup option only in supported MATLAB version which is after R2015a


src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -52,6 +53,7 @@
     private TestRunTypeList testRunTypeList;
     private String matlabRoot;
     private EnvVars env;
+    private FilePath nodeSpecificMatlabRoot;
```
acampbel: What was the motivation for moving this into a property. I see it is determined now within perform as opposed to execMatlabCommand, but curious what motivated the change.

src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -554,10 +556,14 @@ private synchronized int execMatlabCommand(FilePath workspace, Launcher launcher
...
-    private String[] getPreRunnerSwitches() {
+    private String[] getPreRunnerSwitches() throws MatlabVersionNotFoundException {
         String[] preRunnerSwitches =
                 {getLocalMatlab() + nodeSpecificfileSeparator + "bin" + nodeSpecificfileSeparator + "matlab", "-nosplash",
-                        "-nodesktop", "-noAppIcon"};
+                        "-nodesktop"};
+        MatlabReleaseInfo rel = new MatlabReleaseInfo(this.nodeSpecificMatlabRoot);
+        if(!rel.verLessThan(MatlabBuilderConstants.BASE_MATLAB_VERSION_NO_APP_ICON_SUPPORT)) {
+        	preRunnerSwitches =  (String[]) ArrayUtils.add(preRunnerSwitches, "-noAppIcon");
+        }
```
mcafaro: As far as I know, VersionInfo.xml did not exist in 8.6 (R2015b). Are you going to update MatlabReleaseInfo to examine Content.m for the version in the event VersionInfo.xml is not available? Or are you going to wait until you switch over to run_matlab_command?

src/main/java/com/mathworks/ci/MatlabBuilder.java
nbhoski: Since MatlabReleaseInfo now accepts the FilPath instead String . and we could construct the FilePath one we get the Launcher. Hence i am creating FilPath Object as soon i get the Launcher and then make it a prperty so that i could use it across all methods instead of passing it as a parameter to the next one . in our case 
execMatlabCommand & getPreRunnerSwitches methods.

src/main/java/com/mathworks/ci/MatlabBuilder.java
nbhoski: Makes sense, I  by default send  VERSION_16B (9.1) if i dont get VersionInfo. I could adopt Content.m and update the MatlabReleaseInfo  however i see the point for waiting till we adopt run_matlab_command. I feel it only depends on how early we could release the changes and how important this fix is. I will have the MatlabRelease Info updated with Content.m logic. We can disscuss about it.

src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -554,10 +556,14 @@ private synchronized int execMatlabCommand(FilePath workspace, Launcher launcher
     }
 
 
-    private String[] getPreRunnerSwitches() {
+    private String[] getPreRunnerSwitches() throws MatlabVersionNotFoundException {
         String[] preRunnerSwitches =
                 {getLocalMatlab() + nodeSpecificfileSeparator + "bin" + nodeSpecificfileSeparator + "matlab", "-nosplash",
-                        "-nodesktop", "-noAppIcon"};
+                        "-nodesktop"};
+        MatlabReleaseInfo rel = new MatlabReleaseInfo(this.nodeSpecificMatlabRoot);
```
mw-akumar: We are creating another MatlabReleaseInfo object here, whereas one was already created in the earlier method. This means that we will parse the same XML file twice. Since MatlabReleaseInfo is not changing we could store this as a class property.

src/main/java/com/mathworks/ci/MatlabBuilder.java

src/main/java/com/mathworks/ci/MatlabBuilder.java

PR #64, opened 2020-01-08
Version search capability from Contents file for version older than 17b

nbhoski:


src/main/java/com/mathworks/ci/MatlabReleaseInfo.java
```diff
@@ -107,7 +114,22 @@ public boolean verLessThan(double version) throws MatlabVersionNotFoundException
...
-                    versionInfoCache.putAll(VERSION_OLDER_THAN_17A);
+                	FilePath contentFile = new FilePath(this.matlabRoot,"toolbox/matlab/general/Contents.m");
+                	InputStream in = contentFile.read();
+                	BufferedReader br = new BufferedReader(new InputStreamReader(in));
+                	
+                	//Skip first line and capture the second line.
+                	br.readLine();
+                	String versionLine = br.readLine();
+                	
+                	String actualVersion = VERSION_16B;
```
acampbel: Id say its better to use a Null here rather than the 16b default. This lets us know that there was a problem as opposed to giving the wrong version

nbhoski: Done.

src/main/java/com/mathworks/ci/MatlabReleaseInfo.java
```diff
@@ -107,14 +106,31 @@ public boolean verLessThan(double version) throws MatlabVersionNotFoundException
                 else if(!this.matlabRoot.exists()){
                     throw new NotDirectoryException("Invalid matlabroot path");
                 }else {
-                    versionInfoCache.putAll(VERSION_OLDER_THAN_17A);
-                }
-                
+					// Get the version information from Contents.m file when VersionInfo.xml is not
+					// present.
+					FilePath contentFile = new FilePath(this.matlabRoot, CONTENTS_FILE);
+					String actualVersion = null;
```
mw-akumar: formatting is still way off.

src/main/java/com/mathworks/ci/MatlabReleaseInfo.java
```diff
@@ -107,14 +106,31 @@ public boolean verLessThan(double version) throws MatlabVersionNotFoundException
...
+					try (InputStream in = contentFile.read();
+							BufferedReader br = new BufferedReader(new InputStreamReader(in, StandardCharsets.UTF_8))) {
```
mw-akumar: As mentioned earlier, MATLAB files will be UTF-8 encoded in 20a or later. In older releases MATLAB m files are encoded in the "default encoding". However MATLAB's notion of "default encoding" may not always match the default encoding that the Java code in the Jenkins plugin will use, and it will be hard to get this information without starting MATLAB.

However, if Contents.m files contain ASCII only content, then we don't have to worry about the encoding here. Hence UTF-8 is a safe choice.

Issue #44, opened 2019-11-15
Cannot add empty TestRunnerPlugin array to TestRunner


mcafaro:
TestRunner.addPlugin does not allow adding an empty TestRunnerPlugin array.

https://github.com/mathworks/jenkins-matlab-plugin/blob/e824674a8f3a5a106ecfcdbad815b69467ebb37d/src/main/resources/com/mathworks/ci/MatlabBuilder/runMatlabTests.m#L40

PR #66, opened 2020-01-13
Fix empty plugin bug

acampbel:
Note: Need access to matlab in CI of this project to test this file in automation. This will instead be addressed by scriptgen infrastructure.


acampbel:
closes #44

PR #70, opened 2020-02-04
Disabled UI elements for old build steps.

nbhoski:
Disabled all UI elements for older build steps.

### DIFF
--- a/src/main/resources/com/mathworks/ci/MatlabBuilder/config.jelly
+++ b/src/main/resources/com/mathworks/ci/MatlabBuilder/config.jelly
@@ -1,7 +1,7 @@
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
 
- <font color="red"> This build step is depricated. Consider moving to the new MATLAB build steps. Current build however, will work with restricted access.</font>   
+ <font color="red">Using this build step is not recommended and will be removed in a feature release. Use “Run MATLAB Tests” or “Run MATLAB Command” instead.</font>   
  <f:section>
 	  <f:entry title="MATLAB root " field="matlabRoot">
 	        <f:textbox disabled="disabled"/>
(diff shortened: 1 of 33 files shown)


src/main/webapp/disable.js
```diff
@@ -0,0 +1,48 @@
...
+disableAllElements(taCoberturaChkBx);
+
+var taModelCoverageChkBx = getElementsByXpath("//input[@name='taModelCoverageChkBx']");
+disableAllElements(taModelCoverageChkBx);
+
+
+//Function to disable all occurrences of given web element.
+function disableAllElements(nodeName){
+  var a =[];
+  a.push(nodeName.iterateNext());
```
mw-akumar: I think the structure of this code can be simplified (sorry for wrong js syntax)


node = nodeName.IterateNext();
while((node  !== null) {
  a.push(node);
 node = nodeName.IterateNext();
}
Thus the array doesn't need a terminating null because size of array is available.
The second while loop should just loop on the number of elements in the array.

src/main/webapp/disable.js
```diff
@@ -0,0 +1,48 @@
...
+  while(a[a.length-1] !== null){
+      a.push(nodeName.iterateNext());
+  }
+
+  var i=0;
+  while(a[i] !== null){
+      a[i].disable();
+      i=i+1;
+  }
```
mcafaro: Are you intending to use 2 or 4 spaces for indentation? Consider using a tool like jslint to enforce consistency.

nbhoski: Thanks for the suggestion Mark . switched to Jslint default for Java script code.

src/main/webapp/disable.js
nbhoski: Done.

src/main/resources/com/mathworks/ci/MatlabBuilder/config.jelly
```diff
@@ -1,7 +1,7 @@
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
 
- <font color="red"> This build step is depricated. Consider moving to the new MATLAB build steps. Current build however, will work with restricted access.</font>   
+ <font color="red">Using this build step is not recommended and will be removed in a feature release. Use “Run MATLAB Tests” or “Run MATLAB Command” instead.</font>
```
nbhoski: @mw-hrastega May i request you to have a look at this warning message

mw-hrastega: Looks fine to me. Thanks.

PR #72, opened 2020-02-19
Separate build steps by adding two new builders 

nbhoski:
This PR includes the following changes.
* Added new build environment instance named "Add MATLAB to PATH"
image
* Matlab root will now be accepted through this.
* Added two new build steps
image
* All new builds for "Run MATLAB test" will automatically get delegated to new builder class
* "Run MATLAB tests " will now look like below
image
* "Run MATLAB Command " will look like below
image
* Moved all command construction-related helper methods to a common utility class.
* Added tests for both new builder 
* updated the existing test for old builders 

Things Not included in the review :
* Refactoring of existing MatlabBuilder class is yet to be done . i shall evaluate the possibility of using new command construct util. i will track this task with an appropriate issue and raise a separate PR for the same.


nbhoski:
The azure check may fail due to static code analyzer issues. I am looking into it.

src/main/resources/com/mathworks/ci/MatlabScriptBuilder/config.jelly
```diff
@@ -0,0 +1,8 @@
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
+    
+	  <f:entry title="MATLAB command" field="matlabCommand">
```
mcafaro: Consider shorting this to "Command". This is right under the title "Run MATLAB Command", so it probably does not need to restate "MATLAB".

src/main/java/com/mathworks/ci/CommandConstructUtil.java
```diff
@@ -0,0 +1,143 @@
+package com.mathworks.ci;
+/*
+ * Copyright 2020-2021 The MathWorks, Inc.
```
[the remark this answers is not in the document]
nbhoski: Done.

src/main/java/com/mathworks/ci/CommandConstructUtil.java
```diff
@@ -0,0 +1,143 @@
+package com.mathworks.ci;
+/*
+ * Copyright 2020-2021 The MathWorks, Inc.
+ * 
+ * MATLAB command construction utility class used to construct the startup command based on the
+ * builders. Author : Nikhil Bhoski email : nbhoski@mathworks.com Date : 11/02/2020
+ */
```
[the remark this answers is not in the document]
nbhoski: Done. using java doc format

src/main/java/com/mathworks/ci/MatlabScriptBuilder.java
```diff
@@ -0,0 +1,146 @@
...
+import hudson.model.AbstractProject;
+import hudson.model.Result;
+import hudson.model.Run;
+import hudson.model.TaskListener;
+import hudson.tasks.BuildStepDescriptor;
+import hudson.tasks.Builder;
+import jenkins.tasks.SimpleBuildStep;
+import net.sf.json.JSONObject;
+
+public class MatlabScriptBuilder extends Builder implements SimpleBuildStep {
```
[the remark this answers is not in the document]
nbhoski: Sure, Picking the names same as build steps 
RunMatlabCommandBuilder

src/main/java/com/mathworks/ci/MatlabTestRunBuilder.java
```diff
@@ -0,0 +1,341 @@
...
+import hudson.model.TaskListener;
+import hudson.tasks.BuildStepDescriptor;
+import hudson.tasks.Builder;
+import hudson.util.FormValidation;
+import hudson.util.FormValidation.Kind;
+import jenkins.model.Jenkins;
+import jenkins.tasks.SimpleBuildStep;
+import net.sf.json.JSONObject;
+
+public class MatlabTestRunBuilder extends Builder implements SimpleBuildStep {
```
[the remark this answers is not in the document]
nbhoski: Picking the names same as build steps 
RunMatlabTestsBuilder

src/main/java/com/mathworks/ci/MatlabBuildWrapper.java
```diff
@@ -0,0 +1,147 @@
...
+import hudson.Launcher;
+import hudson.model.AbstractProject;
+import hudson.model.Run;
+import hudson.model.TaskListener;
+import hudson.tasks.BuildWrapperDescriptor;
+import hudson.util.FormValidation;
+import hudson.util.FormValidation.Kind;
+import jenkins.tasks.SimpleBuildWrapper;
+
+public class MatlabBuildWrapper extends SimpleBuildWrapper {
```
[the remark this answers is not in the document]
nbhoski: Changed it to AddMatlabToPathBuildWrapper

src/main/java/com/mathworks/ci/MatlabScriptBuilder.java
```diff
@@ -0,0 +1,146 @@
...
+    public void perform(@Nonnull Run<?, ?> build, @Nonnull FilePath workspace,
+            @Nonnull Launcher launcher, @Nonnull TaskListener listener)
+            throws InterruptedException, IOException {
+
+        // Set the environment variable specific to the this build
+        setEnv(build.getEnvironment(listener));
+
+        // Create command util for command constrcution.
+        String matlabRoot = this.env.get("matlabroot");
+        cmdUtils = new CommandConstructUtil(launcher, matlabRoot);
```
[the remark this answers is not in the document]
nbhoski: My initial implementation was to call MATLAB dumbly. with a caveat . due to Java ProcessLaunchers limitations (known currently to me) the default shell process would not get hold of the Environment variables that we set in the wrapper class. it could be only accessed by the child process. hence to make it work i had to invoke the command similar to below
/bin/bash -c "matlab -batch" (on Linux)
cmd.exe /C matlab -batch (on Windows)
both work fine with -batch option however when it comes to -r option the command construction cribs about parenthesis & spaces. It also requires to escape all the "(" , ")" & Single quotes which looks like below.
image

Currently, I am investigating this issue and checking how we could resolve it and make the command look as clean as possible (Without any escape characters &  prefixing shell program executables. this issue was causing delays to out review process hence implemented it in this way for time being. I shall raise new PR once i get to convincing solution for the same.

src/main/resources/com/mathworks/ci/MatlabScriptBuilder/help-matlabCommand.html
```diff
@@ -0,0 +1,9 @@
+<div>
+ To run your tests, enter MATLAB commands separated by semicolons.<br>
+ <b>Example:</b> cd c:\MATLAB\tests; runtests('IncludingSubfolder','true');<br>
```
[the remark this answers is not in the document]
nbhoski: @mw-hrastega  please suggest.

src/main/resources/com/mathworks/ci/MatlabScriptBuilder/config.jelly
nbhoski: Thats a good point, however I feel Command will become very generic. matlabCommand is rather more aligned to the UI elements . and it could be probably easier to co related the WebElements in general as well as for automation needs in future.

mcafaro: I think it is fine if the underlying field name stays "matlabCommand". I'm more interested in the "MATLAB command" title that I think would be better as "Command".

74829958-24dda180-5338-11ea-8555-1bc03ca9f830

Perhaps others can weigh in.

src/main/java/com/mathworks/ci/MatlabScriptBuilder.java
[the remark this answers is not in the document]
mcafaro: Perhaps the use of `run_matlab_command` would help. Not sure if you're ready to switch over yet but it is available at:

https://ssd.mathworks.com/supportfiles/ci/run-matlab-command/v0/run-matlab-command.zip

You could invoke it like you're doing with `-batch` and it should take care of things for you:

Linux:
`/bin/bash -c "run_matlab_command.sh runMatlabTests(...)"`

Windows:
`cmd.exe /C run_matlab_command.bat runMatlabTests(...)`

src/main/resources/com/mathworks/ci/MatlabScriptBuilder/help-matlabCommand.html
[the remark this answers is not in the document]
mw-hrastega: Any thoughts about this?

Enter the semicolon-separated commands you want to run in MATLAB. If a command represents a MATLAB function or script, do not specify the file extension.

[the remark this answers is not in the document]
mcafaro: The "semicolon-separated" comment might be unnecessary or, at least, relegated to a side-note. Perhaps:

Enter the command you want to run in MATLAB. If the command represents a MATLAB function or script, do not specify the file extension.

src/main/resources/com/mathworks/ci/MatlabScriptBuilder/config.jelly
mw-hrastega: I am in favor of Run MATLAB Command-->Command.

src/main/java/com/mathworks/ci/MatlabScriptBuilder.java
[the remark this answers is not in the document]
acampbel: If it is easy to incorporate I like the idea of using run_matlab_command, but we'd probably need to include it with the Jenkins Plugin as a build step, grabbing the "released" version from the location Mark references above. 

I also think that it would be good to incorporate the script generator, but perhaps that is something that can wait until later? It seems like if starting with run_matlab_command helps resolve this problem and it doesn't consistute a large change in scope then it ay be worth doing now.

src/main/resources/com/mathworks/ci/MatlabScriptBuilder/config.jelly
nbhoski: Done.

src/main/resources/com/mathworks/ci/MatlabScriptBuilder/help-matlabCommand.html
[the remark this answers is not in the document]
nbhoski: Done. Added a separate example fro command and script.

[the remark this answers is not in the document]
nbhoski: Done. Added a separate example fro command and script.

src/main/java/com/mathworks/ci/MatlabScriptBuilder.java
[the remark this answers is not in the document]
nbhoski: I think its doable. lot of my code will be reduced with this. working on it to adopt the changes . need bit of testing. i will update the review request once done.

src/main/resources/com/mathworks/ci/RunMatlabCommandBuilder/help-matlabCommand.html
```diff
@@ -1,7 +1,8 @@
 <div>
- To run your tests, enter MATLAB commands separated by semicolons.<br>
- <b>Example:</b> cd c:\MATLAB\tests; runtests('IncludingSubfolder','true');<br>
- 
+ Enter the command you want to run in MATLAB. If the command represents a MATLAB function or script, do not specify the file extension.<br>
+ <b>Examples:</b><br>
+ <b>To run MATLAB commands :</b> cd c:\MATLAB\tests; runtests('IncludingSubfolder','true')<br> 
+ <b>To run MATLAB function or a script file :</b> cd c:\MATLAB\myscripts; runMyScript(arg1, arg2...)<br>
```
mw-hrastega: Scripts or functions: 
cd C:\MATLAB\myMATLABCode; runMyScript; runMyFunction(arg1,arg2);

src/main/resources/com/mathworks/ci/RunMatlabCommandBuilder/help-matlabCommand.html
```diff
@@ -1,7 +1,8 @@
 <div>
- To run your tests, enter MATLAB commands separated by semicolons.<br>
- <b>Example:</b> cd c:\MATLAB\tests; runtests('IncludingSubfolder','true');<br>
- 
+ Enter the command you want to run in MATLAB. If the command represents a MATLAB function or script, do not specify the file extension.<br>
+ <b>Examples:</b><br>
+ <b>To run MATLAB commands :</b> cd c:\MATLAB\tests; runtests('IncludingSubfolder','true')<br>
```
mw-hrastega: Commands: 
cd c:\MATLAB\tests; runtests('IncludingSubfolder','true')

src/main/java/com/mathworks/ci/MatlabScriptBuilder.java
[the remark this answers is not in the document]
acampbel: I think copying the zip file, putting it into resources, and then including it in the jar/hpi is the right approach, but I would say we should not check in those files but instead do it as part of the build.

src/main/resources/com/mathworks/ci/RunMatlabCommandBuilder/help-matlabCommand.html
nbhoski: Done.

src/main/resources/com/mathworks/ci/RunMatlabCommandBuilder/help-matlabCommand.html
nbhoski: Done.

src/main/java/com/mathworks/ci/MatlabScriptBuilder.java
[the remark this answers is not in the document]
nbhoski: * I have made some changes to pom file to resolve the dependancy run time and package it.
i am making dure that the zip file is unpacked and copied under reources folder and then it will be packaged in .jar  
* Adopted Mark's changes.
* both builder classes are changed accordingly.
* Run tests works perfectly fine on both Windows / Linux 
* Run command works perfectly in Windows.

* We still have an issue with /bin/bash -c option when  it comes to running Custom command s. i still need to escape for commands like disp() and seems to be an issue still to identify commands with parenthesis and single quotes even if we escape them as shown below.
image

image

src/main/resources/com/mathworks/ci/RunMatlabCommandBuilder/help-matlabCommand.html
mw-hrastega: I am not sure why I can't see the change. However, here are my concerns:

1) How can a script accept input arguments?
2) We typically say run a function or script, but not run a "file".
3) You say "run MATLAB function or a script". The article "a" should be used consistently. (It should also appear before MATLAB)

I would rewrite the entire statement as follows:
<b>Run a function :</b> cd C:\MATLAB\myMATLABFiles; runMyFunction(arg1,arg2)<br>

src/main/resources/com/mathworks/ci/RunMatlabCommandBuilder/help-matlabCommand.html
mw-hrastega: In line with my comment below, I would rewrite this statement as follows:

<b>Run MATLAB commands :</b> cd C:\MATLAB\tests; runtests('IncludingSubfolder','true')<br>

src/main/java/com/mathworks/ci/MatlabScriptBuilder.java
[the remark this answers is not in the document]
nbhoski: This one is resolved now. since the restriction of using /bin/bash was due to availability of PATH variable changes only to child process. run_matlab_command script itself acts as child process and PATH will be available to the script. so we now no need to escape the path and use /bin/bash 

image

src/main/resources/com/mathworks/ci/RunMatlabCommandBuilder/help-matlabCommand.html
nbhoski: Done.

src/main/java/com/mathworks/ci/AddMatlabToPathBuildWrapper.java
```diff
@@ -0,0 +1,155 @@
...
+        return this.env == null ? getMatlabRootFolder() : this.env.expand(getMatlabRootFolder());
+    }
+
+    private void setEnv(EnvVars env) {
+        this.env = env;
+    }
+
+    @Symbol("Matlab")
+    @Extension
+    public static final class MatlabBuildWrapperDescriptor extends BuildWrapperDescriptor {
```
mcafaro: Descriptor class name needs to be updated to match the new Wrapper name.

src/main/java/com/mathworks/ci/CommandConstructUtil.java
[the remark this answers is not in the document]
mcafaro: Doesn't look like this changed.

src/main/java/com/mathworks/ci/CommandConstructUtil.java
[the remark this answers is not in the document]
mcafaro: Doesn't look like this changed.

src/main/java/com/mathworks/ci/RunMatlabCommandBuilder.java
```diff
@@ -0,0 +1,137 @@
...
+        this.env = env;
+    }
+
+    private EnvVars getEnv() {
+        return this.env;
+    }
+
+    @Symbol("RunMatlabCommand")
+    @Extension
+    public static class MatlabScriptDescriptor extends BuildStepDescriptor<Builder> {
```
mcafaro: Descriptor class name needs to be updated to match the new Wrapper name.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -0,0 +1,340 @@
...
+        this.env = env;
+    }
+    
+    private EnvVars getEnv() {
+        return this.env;
+    }
+    
+    @Symbol("RunMatlabTests")
+    @Extension
+    public static class MatlabTestDescriptor extends BuildStepDescriptor<Builder> {
```
mcafaro: Descriptor class name needs to be updated to match the new Wrapper name.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -0,0 +1,340 @@
...
+            String matlabRunnerTarget, FilePath targetWorkspace)
+            throws IOException, InterruptedException {
+        final ClassLoader classLoader = getClass().getClassLoader();
+        FilePath targetFile =
+                new FilePath(targetWorkspace, Message.getValue(matlabRunnerTarget));
+        InputStream in = classLoader.getResourceAsStream(matlabRunnerResourcePath);
+        targetFile.copyFrom(in);
+        //set executable permission to the file.
+        targetFile.chmod(0777);
+    }*/
```
mcafaro: Can be removed from code

src/main/java/com/mathworks/ci/RunMatlabCommandBuilder.java
```diff
@@ -0,0 +1,137 @@
...
+                launcher = launcher.decorateByPrefix("cmd.exe","/C");
+                matlabLauncher = launcher.launch().pwd(workspace).envs(envVars).cmds("run_matlab_command.bat","\""+cmdUtils.constructCommandForRunCommand(getCommand())+"\"").stdout(listener);
+                //Copy runner.bat for Windows platform in workspace.
+                cmdUtils.copyMatlabScratchFileInWorkspace(MatlabBuilderConstants.BAT_RUNNER_SCRIPT, "Builder.matlab.runner.script.target.file.windows.name", targetWorkspace);
+            }
+        } catch (Exception e) {
+            listener.getLogger().println(e.getMessage());
+            return 1;
+        }
+        return matlabLauncher.join();
```
mcafaro: Consider refactoring. It looks like a large portion of this code block is repeated in `RunMatlabCommandBuilder` and `RunMatlabTestsBuilder`.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -0,0 +1,340 @@
...
+        ProcStarter matlabLauncher;
+        try {
+            // Get matlabroot set in wrapper class.
+            String matlabRoot = envVars.get("matlabroot");     
+            CommandConstructUtil cmdUtils = new CommandConstructUtil(launcher, matlabRoot);
+            matlabLauncher = launcher.launch().pwd(workspace).envs(envVars);
+            FilePath targetWorkspace = new FilePath(launcher.getChannel(), workspace.getRemote());
+            if(launcher.isUnix()) {
+                matlabLauncher = launcher.launch().pwd(workspace).envs(envVars).cmds("/bin/bash","-c","./run_matlab_command.sh "+cmdUtils.constructCommandForTest(getInputArguments())).stdout(listener);
+                //Copy runner .sh for linux platform in workspace.
```
mcafaro: I don't think we need to copy `run_matlab_command` to the workspace. I don't believe a user will need to be looking at this file. Can you use a temp folder instead?

nbhoski: I can try using temp. but i guess it might need a round of discussion about where to create temp folder both for Linux & for Windows . also we need to provide runMatlabtest with an absolute path which would make the whole command look bit odd (not any of the concern though) . we could also explore the option of deleteing it from workspace after build completes.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
nbhoski: Done. wrong commit ID

src/main/java/com/mathworks/ci/AddMatlabToPathBuildWrapper.java
nbhoski: Done.

src/main/java/com/mathworks/ci/RunMatlabCommandBuilder.java
nbhoski: Done

src/main/java/com/mathworks/ci/RunMatlabCommandBuilder.java
nbhoski: Done. added new Interface with default method implementations and both builders are now implementing this interface.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
nbhoski: BATCH & SHELL uses 
**java.io.tmpdrive** system path to create a scratch file and they invoke it using the absolute path. as shown below 
image

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
nbhoski: Done.

src/main/java/com/mathworks/ci/MatlabBuild.java
```diff
@@ -0,0 +1,68 @@
...
+     * @param launcher
+     * @param listener
+     * @param envVars
+     * @param matlabCommand
+     * @return
+     * @throws IOException
+     * @throws InterruptedException
+     */
+    default ProcStarter getProcessToRunMatlabCommand(FilePath workspace, Launcher launcher,TaskListener listener, EnvVars envVars, String matlabCommand) throws IOException, InterruptedException {
+        ProcStarter matlabLauncher;
```
mw-akumar: Is this Java convention to declare first and then assign? Or could this just be
ProcStarter matlabLauncher = launher.launch().....

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -0,0 +1,321 @@
...
+                } catch (MatlabVersionNotFoundException e) {
+                    return FormValidation.warning(Message.getValue("Builder.invalid.matlab.root.warning"));
+                }
+            }
+            
+            
+            return FormValidation.ok();
+        };
+        
+        public FormValidation getFirstErrorOrWarning(
```
mw-akumar: There are 2 different definitions of getFirstErrorOrWarning. Should this be in a common place? And why would the implementations differ a bit?

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -0,0 +1,321 @@
...
+            }catch (Exception e) {
+                return FormValidation.warning(Message.getValue("Builder.invalid.matlab.root.warning"));
+            }
+           
+            return FormValidation.ok();
+        }
+    }
+
+    @Override
+    public void perform(@Nonnull Run<?, ?> build, @Nonnull FilePath workspace,
```
mw-akumar: Both perform methods have the exact same implementation. Is the duplication unavoidable?

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -0,0 +1,321 @@
...
+            return 1;
+        }
+        return matlabLauncher.join();
+    }
+    
+    public String constructCommandForTest(String inputArguments) {
+        String runCommand;
+        String matlabFunctionName = FilenameUtils.removeExtension(
+                Message.getValue(MatlabBuilderConstants.MATLAB_RUNNER_TARGET_FILE));
+        runCommand = "exit(" + matlabFunctionName + "(" + inputArguments + "))";
```
mw-akumar: String runCommand = ?

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -0,0 +1,321 @@
...
+        return runCommand;
+    }   
+    
+    // Concatenate the input arguments
+    private String getInputArguments() {
+        String pdfReport = MatlabBuilderConstants.PDF_REPORT + "," + this.getPdfReportChkBx();
```
mw-akumar: here and elsewhere, should "final" be used wherever possible? In C++ we always use 'const' unless we don't want a const.

src/main/java/com/mathworks/ci/AddMatlabToPathBuildWrapper.java
```diff
@@ -0,0 +1,162 @@
...
+
+public class AddMatlabToPathBuildWrapper extends SimpleBuildWrapper {
+
+    private String matlabRootFolder;
+    private EnvVars env;
+
+    @DataBoundConstructor
+    public AddMatlabToPathBuildWrapper() {}
+
+    public String getMatlabRootFolder() {
```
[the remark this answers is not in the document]
nbhoski: earlier matlabRoot was FilePath beacuse we wanted to actually represent a file path to read versionInfo file. now all we do is accept the MATLAB root as string and add it to the node specific EnvVar object. Since we need not to actually navigate to the actual path for any operation we can keep it a single string.

src/main/java/com/mathworks/ci/MatlabBuild.java
nbhoski: the matlabLauncher is updated based on the branching condition for either Linux/Windows hence i choose to keep it out of the branching scope and return it after branching is done.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
nbhoski: This is because our execMatlabCommand() does differ for both the Builders. and build status does depends on what is the result of ExecMatlabCommand() hence  only step setEnv()  becomes redundant. also perform steps are pretty much builder specific even if it is redundant sometimes.

src/main/java/com/mathworks/ci/MatlabBuild.java
nbhoski: Made appropriate changes

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
nbhoski: Done.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
nbhoski: Done for all the identified methods

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
nbhoski: As discussed. moved the implementation of getFirstErrorOrWarning to new utility class. and made it consistent to take only one argument that is lambda function. i tried making matlabRoot as class property and passing it to all the methods however i was facing an  issue with the same. there seem to be an issue with lambda functions if unhanddled exception occurs https://github.com/google/guice/issues/757 it works when i keep matlabroot under try catch.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
nbhoski: Hi Mark i have copied the runner file under tmp folder basically under system property "java.io.tmpdir"  and deleteing the file after build completion . this seems to be the common practice with batch and shell scripts files too. let me know if thinks this is ok.

src/main/resources/com/mathworks/ci/UseMatlabVersionBuildWrapper/help-matlabRootFolder.html
```diff
@@ -0,0 +1,8 @@
+<div>
+  Enter the full path to the MATLAB root folder, which is returned by the <i><b>matlabroot</b></i> function. The specified MATLAB will be added to PATH variable and will be used across this build.
+  If this field is empty then default matlabroot from PATH variable will be picked if added. <br><br><br>Example:<br><br><i><b><u>Windows:</u></b></i><br> C:\Program Files\MATLAB\R2019a <br>
```
mw-hrastega: I would try to reword these sentences:

"The specified MATLAB will be added to PATH variable and will be used across this build.
If this field is empty then default matlabroot from PATH variable will be picked if added."

as

"The specified directory will be added to the front of the PATH environment variable, and the specified MATLAB version will be used to perform the build.
You must not leave this field blank if no MATLAB directory exists on the path. If you leave the field blank, then the system tries to use the MATLAB directory closest to the top of the system path."

Does it make sense?

nbhoski: Yes its clear , however i have only one word suggestion let me know if we could use prefixed instead added to the front. which looks as below.

The specified directory will be prefixed to the  PATH environment variable, and the specified MATLAB version will be used to perform the build.
You must not leave this field blank if no MATLAB directory exists on the path. If you leave the field blank, then the system tries to use the MATLAB directory closest to the top of the system path.

I am making the changes as per your suggestion for review however if you agree on my suggestion i shall change it else will keep it as is and consider it approved.

mw-hrastega: Hi Nikhil,

Thanks for your suggestion. I think we can use "prepended" instead of "prefixed". So the text would become:

The specified directory will be prepended to the PATH environment variable, and the specified MATLAB version will be used to perform the build.
You must not leave this field blank if no MATLAB directory exists on the path. If you leave the field blank, then the system tries to use the MATLAB directory closest to the top of the system path.

Thanks,
Houman

mw-hrastega:
The changes look fine to me.

src/main/resources/com/mathworks/ci/UseMatlabVersionBuildWrapper/help-matlabRootFolder.html
nbhoski: Done.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
nbhoski: @mcafaro  the changes are in already. you may have your review comments.

src/main/resources/com/mathworks/ci/RunMatlabCommandBuilder/help-matlabCommand.html
```diff
@@ -0,0 +1,10 @@
+<div>
+ Enter the command you want to run in MATLAB. If the command represents a MATLAB function or script, do not specify the file extension.<br>
+ <b>Examples:</b><br>
+ <b>Run Commands:</b> results = runtests(‘IncludingSubfolders’,true); assert(all(~[results.Failed]))<br>
+ <b>Run a script:</b> runMyScript<br>
```
mcafaro: Use consistent capitalization.

Suggest just simplifying this to:
```
Example: myscript
Example: results = runtests(‘IncludingSubfolders’,true); assert(all(~[results.Failed]))
```

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/help-coberturaChkBx.html
```diff
@@ -0,0 +1,5 @@
+<div>
+<br>
+Generate a Cobertura code coverage report, named <b> cobertura.xml, </b> to be stored in the<b> matlabTestArtifacts</b> folder of the current Jenkins workspace.
```
mcafaro: Remove extraneous spaces around `cobertura.xml,`

`named <b> cobertura.xml, </b> to` -> `named <b>cobertura.xml</b>, to`

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/help-junitChkBx.html
```diff
@@ -0,0 +1,5 @@
+<div>
+<br>
+Generate a JUnit style test report, named <b> junittestresults.xml, </b> to be stored in the<b> matlabTestArtifacts</b> folder of the current Jenkins workspace.
```
mcafaro: Remove extraneous spaces around `junittestresults.xml,`

`named <b> junittestresults.xml, </b> to` -> `named <b>junittestresults.xml</b>, to`

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/help-tapChkBx.html
```diff
@@ -0,0 +1,5 @@
+<div>
+<br>
+Generate a TAP style test report, named <b> taptestresults.tap, </b> to be stored in the<b> matlabTestArtifacts</b> folder of the current Jenkins workspace.
```
mcafaro: Remove extraneous spaces around `taptestresults.tap,`

`named <b> taptestresults.tap, </b> to` -> `named <b>taptestresults.tap</b>, to`

src/main/resources/com/mathworks/ci/RunMatlabCommandBuilder/help-matlabCommand.html
nbhoski: Changed to following. I think being explicit about examples for command and script is good idea  
Run command : results = runtests(‘IncludingSubfolders’,true); assert(all(~[results.Failed]))
Run a script : runMyScript

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/help-coberturaChkBx.html
nbhoski: Done.

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/help-junitChkBx.html
nbhoski: done

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/help-tapChkBx.html
nbhoski: Done.

PR #75, opened 2020-03-11
Fix for matrix issue 

nbhoski:
Matrix builds were failing due to synchronization issue . since mtalb_runner script was used by both builds with same name and same location .  the namespacing issue caused accidental deletion of the runner script by by different thread than it had create. Resolved this issue by letting each build step to copy its own file in tmp folder and delete the same file. File uniqueness is mainatined by adding a timeStamp to the file name.

### DIFF
--- a/src/main/java/com/mathworks/ci/MatlabBuild.java
+++ b/src/main/java/com/mathworks/ci/MatlabBuild.java
@@ -33,7 +33,7 @@ public interface MatlabBuild {
             TaskListener listener, EnvVars envVars, String matlabCommand, String uniqueName)
             throws IOException, InterruptedException {
         // Get node specific tmp directory to copy matlab runner script
-        String tmpDir = getNodeSpecificTmpFolderPath();
+        String tmpDir = getNodeSpecificTmpFolderPath(workspace);
         FilePath targetWorkspace = new FilePath(launcher.getChannel(), tmpDir);
         ProcStarter matlabLauncher;
         if (launcher.isUnix()) {
@@ -70,9 +70,12 @@ public interface MatlabBuild {
         targetFilePath.chmod(0755);
     }
 
-    default FilePath getFilePathForUniqueFolder(Launcher launcher, String uniqueName)
+    default FilePath getFilePathForUniqueFolder(Launcher launcher, String uniqueName,FilePath workspace)
             throws IOException, InterruptedException {
-        Computer cmp = Computer.currentComputer();
+        /*Use of Computer is not recommended as jenkins hygeine for pipeline support
+         * https://javadoc.jenkins-ci.org/jenkins/tasks/SimpleBuildStep.html */
+        
+        Computer cmp = workspace.toComputer();
         String tmpDir = (String) cmp.getSystemProperties().get("java.io.tmpdir");
         if (launcher.isUnix()) {
             tmpDir = tmpDir + "/" + uniqueName;
@@ -82,8 +85,8 @@ public interface MatlabBuild {
         return new FilePath(launcher.getChannel(), tmpDir);
     }
 
-    default String getNodeSpecificTmpFolderPath() throws IOException, InterruptedException {
-        Computer cmp = Computer.currentComputer();
+    default String getNodeSpecificTmpFolderPath(FilePath workspace) throws IOException, InterruptedException {
+        Computer cmp = workspace.toComputer();
         String tmpDir = (String) cmp.getSystemProperties().get("java.io.tmpdir");
         return tmpDir;
     }
--- a/src/main/java/com/mathworks/ci/MatlabBuild.java
+++ b/src/main/java/com/mathworks/ci/MatlabBuild.java
@@ -2,15 +2,13 @@ package com.mathworks.ci;
 /**
  * Copyright 2019-2020 The MathWorks, Inc.
  * 
- * Build Interface has two default methods. MATLAB builders can override the 
- * default behavior.
+ * Build Interface has two default methods. MATLAB builders can override the default behavior.
  * 
  */
 
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.UUID;
-
 import hudson.EnvVars;
 import hudson.FilePath;
 import hudson.Launcher;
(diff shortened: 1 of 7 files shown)


nbhoski:
> Rather than add a timestamp to the file, can you create a new tempdir for each iteration, and rather than using timestamp I would suggest using java.util.UUID for a sub-folder name and still use run_matlab_command.sh without modifying the name of the script itself.
> 
> Also, do you know how you can add a test that would have caught this?

Done . Added Matrix build dependency to add Matrix related tests.

src/main/java/com/mathworks/ci/MatlabBuild.java
```diff
@@ -28,46 +30,47 @@
...
         InputStream in = classLoader.getResourceAsStream(sourceFile);
         targetFilePath.copyFrom(in);
         // set executable permission
-        targetFilePath.chmod(0755);
-        
+        targetFilePath.chmod(0755);       
     }
     
-    default FilePath getNodeSpecificMatlabRunnerScript(Launcher launcher) throws IOException, InterruptedException {
+    default FilePath getNodeSpecificMatlabRunnerScript(Launcher launcher,String uniqueName) throws IOException, InterruptedException {
```
mcafaro: Seems like this method name should change to reflect this is now returning a folder and not a script path.

src/main/java/com/mathworks/ci/RunMatlabCommandBuilder.java
```diff
@@ -108,24 +108,25 @@ public void perform(@Nonnull Run<?, ?> build, @Nonnull FilePath workspace,
...
-            matlabLauncher = getProcessToRunMatlabCommand(workspace, launcher, listener, envVars,getCommand());
+            matlabLauncher = getProcessToRunMatlabCommand(workspace, launcher, listener, envVars,getCommand(),uniqueTmpFldrName);
+            return matlabLauncher.join();
         } catch (Exception e) {
             listener.getLogger().println(e.getMessage());
             return 1;
+        }finally {
+        	// Cleanup the runner File from tmp directory
+        	FilePath matlabRunnerScript = getNodeSpecificMatlabRunnerScript(launcher,uniqueTmpFldrName);
+            if(matlabRunnerScript.isDirectory()) {
```
mcafaro: When would this not be a directory?

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -269,32 +268,33 @@ public void perform(@Nonnull Run<?, ?> build, @Nonnull FilePath workspace,
...
             copyFileInWorkspace(MatlabBuilderConstants.MATLAB_RUNNER_RESOURCE,
                     MatlabBuilderConstants.MATLAB_RUNNER_TARGET_FILE, targetWorkspace);
+            return matlabLauncher.join();
         } catch (Exception e) {
             listener.getLogger().println(e.getMessage());
             return 1;
+        }finally {
+            // Cleanup the runner File from tmp directory
+            FilePath matlabRunnerScript = getNodeSpecificMatlabRunnerScript(launcher,uniqueTmpFldrName);
+            if (matlabRunnerScript.isDirectory()) {
```
mcafaro: When would this not be a directory?

src/main/java/com/mathworks/ci/MatlabBuild.java
```diff
@@ -28,46 +30,47 @@
...
         Computer cmp = Computer.currentComputer();
         String tmpDir = (String) cmp.getSystemProperties().get("java.io.tmpdir");
         if(launcher.isUnix()) {
-            tmpDir = tmpDir+"/run_matlab_command.sh";
+            tmpDir = tmpDir+"/"+uniqueName;
         }else {
-            tmpDir = tmpDir+"\\"+"run_matlab_command.bat";
+            tmpDir = tmpDir+"\\"+uniqueName;
         }
```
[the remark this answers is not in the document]
nbhoski: File.separator will be master node specific. if Jenkins instance is running on Windows but the build is running n Linux node the File.separator will always yield the value of master node . hence Launcher is always used to support build running on remote nodes.

src/main/java/com/mathworks/ci/MatlabBuild.java
nbhoski: Done.

src/main/java/com/mathworks/ci/RunMatlabCommandBuilder.java
nbhoski: good Point :) Fixed

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
nbhoski: Done.

src/main/java/com/mathworks/ci/MatlabBuild.java
```diff
@@ -28,46 +30,47 @@
...
-    default ProcStarter getProcessToRunMatlabCommand(FilePath workspace, Launcher launcher,TaskListener listener, EnvVars envVars, String matlabCommand) throws IOException, InterruptedException {
+    default ProcStarter getProcessToRunMatlabCommand(FilePath workspace, Launcher launcher,TaskListener listener, EnvVars envVars, String matlabCommand,String uniqueName) throws IOException, InterruptedException {
         //Get node specific tmp directory to copy matlab runner script
         String tmpDir = getNodeSpecificTmpFolderPath();
         FilePath targetWorkspace = new FilePath(launcher.getChannel(), tmpDir);
         ProcStarter matlabLauncher;
         if(launcher.isUnix()) {
-            matlabLauncher = launcher.launch().pwd(workspace).envs(envVars).cmds(tmpDir+"/run_matlab_command.sh",matlabCommand).stdout(listener);
+        	final String runnerScriptName = uniqueName+"/run_matlab_command.sh";
+            matlabLauncher = launcher.launch().pwd(workspace).envs(envVars).cmds(tmpDir+"/"+runnerScriptName,matlabCommand).stdout(listener);
```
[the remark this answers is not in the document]
nbhoski: I use Google code style. Thanks for noticing this . its weird that formating looks fine in the eclipse but it gets changed on Github sometimes. i had noticed this earlier too.  below is my IDE snapshot 
image
I reapplied the styler format and now it should be fine.

PR #77, opened 2020-03-16
fixed findbug issue and Scratch file copy issue.

nbhoski:
Fixed the Findbug issue . suppressed the warning as workspace is always passed  through perform() method of builder which will always have workspace available. 

Also there is fix for  matlab scratch file copy issue due to new changes in copyMethod () 
test to check the if scratch file is copied in workspace was missing from new RunMatlabTestsBuilder class which is added now.

### DIFF
--- a/src/main/java/com/mathworks/ci/MatlabBuild.java
+++ b/src/main/java/com/mathworks/ci/MatlabBuild.java
@@ -9,8 +9,6 @@ package com.mathworks.ci;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.UUID;
-import javax.annotation.Nonnull;
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.EnvVars;
 import hudson.FilePath;
 import hudson.Launcher;
@@ -72,15 +70,12 @@ public interface MatlabBuild {
         targetFilePath.chmod(0755);
     }
 
-    @SuppressFBWarnings(value = "NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE",
-            justification = "workspace is passed through perform method, which will always have NonNull FilePath")
     default FilePath getFilePathForUniqueFolder(Launcher launcher, String uniqueName, FilePath workspace)
             throws IOException, InterruptedException {
         /*Use of Computer is not recommended as jenkins hygeine for pipeline support
          * https://javadoc.jenkins-ci.org/jenkins/tasks/SimpleBuildStep.html */
         
-        Computer cmp = workspace.toComputer();
-        String tmpDir = (String) cmp.getSystemProperties().get("java.io.tmpdir");
+        String tmpDir = getNodeSpecificTmpFolderPath(workspace);
         if (launcher.isUnix()) {
             tmpDir = tmpDir + "/" + uniqueName;
         } else {
@@ -89,10 +84,11 @@ public interface MatlabBuild {
         return new FilePath(launcher.getChannel(), tmpDir);
     }
 
-    @SuppressFBWarnings(value = "NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE",
-            justification = "workspace is passed through perform method, which will always have NonNull FilePath")
     default String getNodeSpecificTmpFolderPath(FilePath workspace) throws IOException, InterruptedException {
         Computer cmp = workspace.toComputer();
+        if (cmp == null) {
+            throw new IOException(Message.getValue("build.workspace.computer.not.found"));
+        }
         String tmpDir = (String) cmp.getSystemProperties().get("java.io.tmpdir");
         return tmpDir;
     }
@@ -100,5 +96,4 @@ public interface MatlabBuild {
     default String getUniqueNameForRunnerFile() {
         return UUID.randomUUID().toString();
     }
-
 }
--- a/src/main/java/com/mathworks/ci/MatlabBuild.java
+++ b/src/main/java/com/mathworks/ci/MatlabBuild.java
@@ -9,6 +9,8 @@ package com.mathworks.ci;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.UUID;
+import javax.annotation.Nonnull;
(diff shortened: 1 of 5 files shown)


nbhoski:
I also noticed that azure pipeline wont get executed for any other branches except master. We should configure to run azure pipeline check for every branch and for merge of every other branches  including master.

src/main/java/com/mathworks/ci/MatlabBuild.java
```diff
@@ -70,7 +72,9 @@ default void copyFileInWorkspace(String sourceFile, String targetFile, FilePath
         targetFilePath.chmod(0755);
     }
 
-    default FilePath getFilePathForUniqueFolder(Launcher launcher, String uniqueName,FilePath workspace)
+    @SuppressFBWarnings(value = "NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE",
+            justification = "workspace is passed through perform method, which will always have NonNull FilePath")
```
mcafaro: Would it be hard to fix this issue rather than suppress it? Could you just do a null check? I tend to prefer suppressing issues only as a last resort.

src/main/java/com/mathworks/ci/MatlabBuilderConstants.java
```diff
@@ -12,6 +12,7 @@
     static final double BASE_MATLAB_VERSION_EXPORTSTMRESULTS_SUPPORT = 9.6;
     
     static final String MATLAB_RUNNER_TARGET_FILE = "Builder.matlab.runner.target.file.name";
+    static final String MATLAB_TESTS_RUNNER_TARGET_FILE = "runMatlabTests.m";
     static final String MATLAB_RUNNER_RESOURCE = "com/mathworks/ci/MatlabBuilder/runMatlabTests.m";
```
mcafaro: It seems like all of these variables contain values that repeat the `runMatlabTests` function name. You may consider refactoring down to 1 location where the `runMatlabTests` literal is written.

nbhoski: Correct the underlying changes are in the actual copy method which is common across all builders . our implementation now is consistent for all new builder classes.   Variable MATLAB_RUNNER_TARGET_FILE is used in our old MatlabBuilder class as well which in near future will be deprecated.

src/main/java/com/mathworks/ci/MatlabBuild.java
nbhoski: I thought of this as well , however my rational was if workspace for any reason is null then build would fail much before it reaches to perform() method of builder class. However i also tried wrapping it around few null checks like Objects.requireNonNull() and few other but it seems there is open issue for this on find bug https://github.com/spotbugs/spotbugs/issues/651 

Current fix with if(cmp ==null) only seem to be working so far.

src/main/resources/config.properties
```diff
@@ -16,4 +16,5 @@ Builder.matlab.modelcoverage.support.warning = To generate a Cobertura model cov
 Builder.matlab.exportstmresults.support.warning = To export Simulink Test Manager results, use MATLAB R2019a or a newer release.
 Builder.matlab.runner.script.target.file.linux.name = run_matlab_command.sh
 Builder.matlab.runner.script.target.file.windows.name = run_matlab_command.bat
+build.workspace.computer.not.found = Unable to get the current computer for this build.
```
nbhoski: adding @mw-hrastega  to approve this exception message

src/main/java/com/mathworks/ci/MatlabBuild.java
```diff
@@ -70,13 +70,12 @@ default void copyFileInWorkspace(String sourceFile, String targetFile, FilePath
...
+    default FilePath getFilePathForUniqueFolder(Launcher launcher, String uniqueName, FilePath workspace)
             throws IOException, InterruptedException {
         /*Use of Computer is not recommended as jenkins hygeine for pipeline support
          * https://javadoc.jenkins-ci.org/jenkins/tasks/SimpleBuildStep.html */
         
-        Computer cmp = workspace.toComputer();
-        String tmpDir = (String) cmp.getSystemProperties().get("java.io.tmpdir");
+        String tmpDir = getNodeSpecificTmpFolderPath(workspace);
         if (launcher.isUnix()) {
             tmpDir = tmpDir + "/" + uniqueName;
```
mw-akumar: I  think we should avoid having our code do this platform specific path separator handling to the extend possible.
hudson.FilePath and java.io.FilePath classes offer the facility of adding to paths and will abstract out the platform differences.

src/main/java/com/mathworks/ci/MatlabBuild.java
```diff
@@ -87,12 +86,14 @@ default FilePath getFilePathForUniqueFolder(Launcher launcher, String uniqueName
 
     default String getNodeSpecificTmpFolderPath(FilePath workspace) throws IOException, InterruptedException {
         Computer cmp = workspace.toComputer();
+        if (cmp == null) {
+            throw new IOException(Message.getValue("build.workspace.computer.not.found"));
```
mw-akumar: In the unlikely case that cmp is null, how will this exception propagate? Will this be converted to an error that will make sense to the end user?  If it is more of an internal error, should we say so to the user.

nbhoski: Yes, currently we are showing a custom message which is **Unable to get the current computer for this build.** subjected to @mw-hrastega  approval

src/main/java/com/mathworks/ci/MatlabBuild.java
nbhoski: Done.

src/main/resources/config.properties
mw-hrastega: If there is no value in having the word "current", I would remove it. 

"Unable to get the computer for this build."

It might be possible to use "access" instead of "get".

"Unable to access the computer for this build."

nbhoski: Done.

PR #78, opened 2020-03-17
Separate build steps documentation

mw-hrastega:
Updated screenshots and text on CONFIGDOC.md and examples/Run-MATLAB-Tests.md.
Removed redundant screenshots.
Refactored CONFIGDOC.md.


examples/Run-MATLAB-Tests.md
```diff
@@ -35,40 +35,40 @@ Create a new project and configure it by following these steps:
...
 
 ![source_control](https://user-images.githubusercontent.com/48831250/71735877-965fa000-2e1d-11ea-95c4-8b9259308e75.png)
 
-5. In the **Build** section of Jenkins, click **Add build step > Run MATLAB Tests**. A new window opens that enables you to configure the Jenkins plugin for MATLAB. Paste the full path to the MATLAB root folder into the **MATLAB root** field; choose the default Automatic test mode from the **Test mode** drop-down list; and select your desired test artifacts to be generated in the project workspace. (The plugin in this example is configured to generate Cobertura code coverage and JUnit test result reports.) For more information on how to configure the Jenkins plugin for MATLAB, see [Jenkins MATLAB Plugin](https://github.com/jenkinsci/matlab-plugin).
+5. In the **Build Environment** section of Jenkins, select **Use MATLAB Version**. Then, enter the full path to the MATLAB root folder in the **MATLAB root** box. Jenkins uses MATLAB at the specified location to run the tests.
```
acampbel: Similar comment about this really just adding to the PATH, and is only needed when agents don't already have MATLAB on the PATH

CONFIGDOC.md
```diff
@@ -1,109 +1,78 @@
...
-3) In the **Run MATLAB Tests** section of the project, use the **VERSION** axis to specify the locations where MATLAB is installed.
+![build_environment_matrix](https://user-images.githubusercontent.com/48831250/76800665-87488700-67aa-11ea-9dbd-3c3ab518afa7.png)
 
-![image](https://user-images.githubusercontent.com/51316279/72131668-ef41b200-33a2-11ea-9f97-d9476ae0431a.png)
+3) Add the required build steps in the **Build** section:
+   *   If you add the [**Run MATLAB Tests**](#run-matlab-tests) build step, select check boxes for the artifacts to be generated in the project workspace.
+   *   If you add the [**Run MATLAB Command**](#run-matlab-command) build step, enter your commands in the **Command** box. You can use the user-defined axes to create your commands. For example:
 
-4) From the **Test mode** drop-down list, select the **Custom** option. Use the second user-defined axis to create your commands and enter them in the **MATLAB command** field. Then, save your settings and run the build. 
+![run_matlab_command_matrix](https://user-images.githubusercontent.com/48831250/76800855-e3131000-67aa-11ea-99c8-f18b811eea6f.png)
```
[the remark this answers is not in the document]
mw-hrastega: Thanks for catching the typo. Done.

CONFIGDOC.md
```diff
@@ -1,109 +1,78 @@
...
+The **Run MATLAB Tests** build step enables you to easily generate different types of test artifacts. To publish the test results, you can use these artifacts with other Jenkins plugins. The selected artifacts are saved in the **matlabTestArtifacts** folder of the Jenkins workspace. If you do not select any of the test artifact check boxes, the folder is not created. However, the tests still run and test failures fail the build. 
+ 
+![run_matlab_tests](https://user-images.githubusercontent.com/48831250/76800820-d2629a00-67aa-11ea-9fa7-c466e8c17f27.png)
+ 
+The **Run MATLAB Tests** build step produces a MATLAB script file named **runMatlabTests.m** in the Jenkins workspace. The plugin uses this file to run the tests and generate the test artifacts. You can review the contents of the script to understand the testing workflow.
+ 
+**Note:** Producing a PDF test report is not currently supported on MacOS platforms.
+ 
+### Run MATLAB Command
+The **Run MATLAB Command** build step enables you to specify MATLAB commands tailored to your specific needs. For example, you can use this build step to customize your test run.
```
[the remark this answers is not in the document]
mw-hrastega: Done.

CONFIGDOC.md
```diff
@@ -67,7 +67,9 @@ You can define several axes in the **Configuration Matrix** section. For example
 
 ![axis_version_testtag](https://user-images.githubusercontent.com/48831250/76800736-aa733680-67aa-11ea-86a7-a984d5795e11.png)
 
-2) In the **Build Environment** section, select **Use MATLAB version** and then specify the locations where MATLAB is installed using the user-defined axis name. In this example, **$VERSION** in the **MATLAB root** will be replaced by one axis value per build iteration.
+2) In the **Build Environment** section, select **Use MATLAB version** and then specify the locations where MATLAB is installed using the user-defined axis name. (You can skip this step if MATLAB has already been added to the path on the build agents.)
```
acampbel: In the case of the matrix build, when looping over different versions of MATLAB I think it is less likely that they will skip this step. Each build agent can only have one matlab on the PATH, and when the matrix build loops over the parameter it will need to add that specific version to the top of the path.

CONFIGDOC.md
```diff
@@ -44,8 +44,8 @@ Test artifacts are not autogenerated if you choose to run tests using this build
 ## Set Up Freestyle Project
 To configure the plugin for a freestyle project, specify the location where MATLAB is installed as well as the required build steps.
 
-1) In the **Build Environment** section of the project configuration window, select **Use MATLAB version**. Then, enter the value returned by the **matlabroot** function in the **MATLAB root** box.
-  
+1) In the **Build Environment** section of the project configuration window, select **Use MATLAB version**. Then, enter the value returned by the **matlabroot** function in the **MATLAB root** box. (You can skip this step if MATLAB has already been added to the path on the build agent.)
```
acampbel: I think having the skip message here is appropriate

CONFIGDOC.md
```diff
@@ -1,18 +1,18 @@
...
+-  [Add MATLAB to System Path](#add-matlab-to-system-path)
 -  [Specify Build Steps](#specify-build-steps)
       - [Run MATLAB Tests](#run-matlab-tests)
       - [Run MATLAB Command](#run-matlab-command)
 -  [Set Up Freestyle Project](#set-up-freestyle-project)
 -  [Set Up Multi-Configuration Project](#set-up-multi-configuration-project)
 
-## Use MATLAB for Build
-Once you install the plugin on Jenkins, the **Use MATLAB version** checkbox is displayed in the **Build Environment** section of the project configuration window. Selecting the checkbox instructs Jenkins to use MATLAB for your build.
+## Add MATLAB to System Path
```
acampbel: I actually prefer the title you had before. Conceptually they are specifeicaying which matlab they want to use, and its in the details that they learn this means adding it to the PATH

CONFIGDOC.md
```diff
@@ -1,109 +1,78 @@
...
-2) In the **Run MATLAB Tests** section of the project, include the user-defined axis name in the **MATLAB root** field to specify the locations where MATLAB is installed. In this example, **$VERSION** will be replaced by one axis value per build step.
+Test artifacts are not autogenerated if you choose to run tests using this build step. You can generate your desired test artifacts by configuring the test runner in the script or function that you invoke from the **Command** box.
 
-![image](https://user-images.githubusercontent.com/51316279/72131571-aee23400-33a2-11ea-8e76-4b36274b9d5b.png)
+## Freestyle Project Setup
+To configure the plugin for a freestyle project, specify the location where MATLAB is installed as well as the required build steps.
 
-You can select the test artifact check boxes when tests run automatically. Once you have made your selections, save your settings and run the build.
+1) In the **Build Environment** section of the project configuration window, select **Use MATLAB Version**. Then, enter the value returned by the **matlabroot** function in the **MATLAB root** box.
+
```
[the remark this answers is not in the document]
mw-hrastega: Done. Also, add a new section to the top of the document.

CONFIGDOC.md
```diff
@@ -1,18 +1,18 @@
...
 
 ![matlab_path](https://user-images.githubusercontent.com/48831250/76883191-1c528b00-6852-11ea-9242-7d143f8a5c46.png)
 
-When you select the **Use MATLAB version** checkbox, you can specify the full path to the MATLAB installation folder that is used for the build. The specified directory in the **MATLAB root** box is prepended to the PATH environment variable, and the specified MATLAB version is used to perform the build. You must not leave the **MATLAB root** box blank if no MATLAB directory exists on the path. If you leave the box blank, then the system tries to use the MATLAB directory closest to the top of the system path.
+Select the **Use MATLAB version** checkbox to specify the full path to the MATLAB root folder that should be used for the build. The specified directory in the **MATLAB root** box is prepended to the PATH environment variable, and the matlab.exe file in the bin subdirectory is used to perform the build. You must not leave the **MATLAB root** box blank if no MATLAB directory exists on the path. If you do not select the checkbox or if you leave the text box blank, then the system tries to use the MATLAB directory closest to the top of the system path.
```
acampbel: This is not quite right. The user specifies the MATLAB root (not including the bin subdir). What happens is the root + bin subdir is the ting prepended to the PATH.

So the "specified directory" is the MATLAB root, but the thing that is added to the PATH is the bin dir under the root that is specified. The takeaway is that they specify "where" their MATLAB is and the right thing is added to the PATH.

Also, this seems too strong:
    _You must not leave the **MATLAB root** box blank if no MATLAB directory exists on the path._

It is OK if they leave the box blank. They should just know that the plugin nis going to call "matlab" and if it can't find it the build will fail.

I like the idea of clarifying here in this section that if the build agent already has the desired MATLAB on the PATH then there is no need to check this box at all. That might even be a nice workflow. Whoever sets up the agents ensures MATLAB is available and people configuring builds don't even have to worry about it.

examples/Run-MATLAB-Tests.md
mw-hrastega: Done. Please let me know if the changes are fine.

CONFIGDOC.md
mw-hrastega: Thanks for clarifying. I removed the parenthetical information.

CONFIGDOC.md
mw-hrastega: Sure. Reverted to the previous title.

CONFIGDOC.md
mw-hrastega: Good points. How about the revised text?

CONFIGDOC.md
```diff
@@ -1,18 +1,20 @@
...
 
-Select the **Use MATLAB version** checkbox to specify the full path to the MATLAB root folder that should be used for the build. The specified directory in the **MATLAB root** box is prepended to the PATH environment variable, and the matlab.exe file in the bin subdirectory is used to perform the build. You must not leave the **MATLAB root** box blank if no MATLAB directory exists on the path. If you do not select the checkbox or if you leave the text box blank, then the system tries to use the MATLAB directory closest to the top of the system path.
+Select the **Use MATLAB version** checkbox to specify the full path to the MATLAB root folder that should be used for the build. The information in the **MATLAB root** box enables the plugin to prepend MATLAB to the PATH environment variable of the build agent. 
+
+If the build agent already has your desired MATLAB on the path, then you are not required to select the **Use MATLAB version** checkbox. When the checkbox is clear, the plugin uses the MATLAB directory closest to the top of the path. The build fails if the plugin cannot find MATLAB on the path.
```
acampbel: This is great. Two small tweaks:

_When the checkbox is clear, the plugin uses the first MATLAB it encounters on the PATH.

 The build fails if MATLAB is not found on the path._

Doesn't have to be phrased  these ways, but for the first one "closest to the top of the path" just sounded a bit off. and the second sentence I just want to clarify because the plugin doesn't really find MATLAB. It just calls, MATLAB, and if the shell doesn't find it it will fail.

mw-hrastega: Thanks, Andy. I tweaked the the sentences. If the text has any issues, please let me know.

nbhoski:
Merging changes on behalf of @mw-hrastega

PR #81, opened 2020-03-20
update note for Mac platforms

mw-hrastega:
Updated the second to last note on README in response to the issues on Mac platforms.
Please review the note.


mw-hrastega:
1) Updated the screenshot for Run MATLAB Command step.
2) Added a note and table for the MATLAB root folder.

PR #86, opened 2020-04-28
Artifact input change

nbhoski:
* This change is to support an extra text box which will appear on click of the check boxes for each test artifacts. which will allow users to store the test artifacts at the custom location as shown below.

image

* Every text box will have default location pre-populated by default.
* Now Parameters to runMatlabTest() file will be specific to the artifact which is selected. unlike list of all the parameters for non selected values. 
* Removed all UI validation against all test artifact check boxes as matlab root is not mandatory for the build.
* Added new unit tests and reduced some invalid test cases.

### DIFF
--- a/src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
+++ b/src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
@@ -1,27 +1,14 @@
 package com.mathworks.ci;
 
-/**
- * Copyright 2019-2020 The MathWorks, Inc.
- * 
- * MATLAB test run builder used to run all MATLAB & Simulink tests automatically and generate
- * selected test artifacts.
- * 
- */
-
-
-import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.function.Function;
 import javax.annotation.Nonnull;
 import org.apache.commons.io.FilenameUtils;
 import org.jenkinsci.Symbol;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.DataBoundSetter;
-import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.StaplerRequest;
-import com.mathworks.ci.UseMatlabVersionBuildWrapper.UseMatlabVersionDescriptor;
 import hudson.EnvVars;
 import hudson.Extension;
 import hudson.FilePath;
@@ -33,8 +20,6 @@ import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.tasks.BuildStepDescriptor;
 import hudson.tasks.Builder;
-import hudson.util.FormValidation;
-import jenkins.model.Jenkins;
 import jenkins.tasks.SimpleBuildStep;
 import net.sf.json.JSONObject;
 
@@ -202,8 +187,6 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
     @Extension
     public static class RunMatlabTestsDescriptor extends BuildStepDescriptor<Builder> {
 
-        MatlabReleaseInfo rel;
-
         // Overridden Method used to show the text under build dropdown
         @Override
         public String getDisplayName() {
@@ -227,116 +210,6 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
         public boolean isApplicable(
                 @SuppressWarnings("rawtypes") Class<? extends AbstractProject> jobtype) {
             return true;
-        }
-
-
-        /*
-         * Validation for Test artifact generator checkBoxes
-         */
-
-        // Get the MATLAB root entered in build wrapper descriptor
-
-
-
-        public FormValidation doCheckCoberturaChkBxFilePath(@QueryParameter String coberturaChkBxFilePath) {
-            List<Function<String, FormValidation>> listOfCheckMethods =
-                    new ArrayList<Function<String, FormValidation>>();
-            if(!coberturaChkBxFilePath.isEmpty()) {
-                listOfCheckMethods.add(chkCoberturaSupport);
-            }
-            return FormValidationUtil.getFirstErrorOrWarning(listOfCheckMethods, getMatlabRoot());
-        }
-
-        Function<String, FormValidation> chkCoberturaSupport = (String matlabRoot) -> {
-            FilePath matlabRootPath = new FilePath(new File(matlabRoot));
-            rel = new MatlabReleaseInfo(matlabRootPath);
-            final MatrixPatternResolver resolver = new MatrixPatternResolver(matlabRoot);
-            if (!resolver.hasVariablePattern()) {
-                try {
-                    if (rel.verLessThan(
-                            MatlabBuilderConstants.BASE_MATLAB_VERSION_COBERTURA_SUPPORT)) {
-                        return FormValidation.warning(
-                                Message.getValue("Builder.matlab.cobertura.support.warning"));
-                    }
-                } catch (MatlabVersionNotFoundException e) {
-                    return FormValidation
-                            .warning(Message.getValue("Builder.invalid.matlab.root.warning"));
-                }
-            }
-
-
-            return FormValidation.ok();
-        };
-                
-        public FormValidation doCheckModelCoverageChkBxFilePath(
-                @QueryParameter String modelCoverageChkBxFilePath) {
-            List<Function<String, FormValidation>> listOfCheckMethods =
-                    new ArrayList<Function<String, FormValidation>>();
-            if (!modelCoverageChkBxFilePath.isEmpty()) {
-                listOfCheckMethods.add(chkModelCoverageSupport);
-            }
-            return FormValidationUtil.getFirstErrorOrWarning(listOfCheckMethods, getMatlabRoot());
-        }
-
-        Function<String, FormValidation> chkModelCoverageSupport = (String matlabRoot) -> {
-            FilePath matlabRootPath = new FilePath(new File(matlabRoot));
-            rel = new MatlabReleaseInfo(matlabRootPath);
-            final MatrixPatternResolver resolver = new MatrixPatternResolver(matlabRoot);
-            if (!resolver.hasVariablePattern()) {
-                try {
-                    if (rel.verLessThan(
-                            MatlabBuilderConstants.BASE_MATLAB_VERSION_MODELCOVERAGE_SUPPORT)) {
-                        return FormValidation.warning(
-                                Message.getValue("Builder.matlab.modelcoverage.support.warning"));
-                    }
-                } catch (MatlabVersionNotFoundException e) {
-                    return FormValidation
-                            .warning(Message.getValue("Builder.invalid.matlab.root.warning"));
-                }
-            }
-
-
-            return FormValidation.ok();
-        };
-
-        public FormValidation doCheckStmResultsChkBxFilePath(@QueryParameter String stmResultsChkBxFilePath) {
-            List<Function<String, FormValidation>> listOfCheckMethods =
-                    new ArrayList<Function<String, FormValidation>>();
-            if (!stmResultsChkBxFilePath.isEmpty()) {
-                listOfCheckMethods.add(chkSTMResultsSupport);
-            }
-            return FormValidationUtil.getFirstErrorOrWarning(listOfCheckMethods, getMatlabRoot());
-        }
-
-        Function<String, FormValidation> chkSTMResultsSupport = (String matlabRoot) -> {
-            FilePath matlabRootPath = new FilePath(new File(matlabRoot));
-            rel = new MatlabReleaseInfo(matlabRootPath);
-            final MatrixPatternResolver resolver = new MatrixPatternResolver(matlabRoot);
-            if (!resolver.hasVariablePattern()) {
-                try {
-                    if (rel.verLessThan(
-                            MatlabBuilderConstants.BASE_MATLAB_VERSION_EXPORTSTMRESULTS_SUPPORT)) {
-                        return FormValidation.warning(Message
-                                .getValue("Builder.matlab.exportstmresults.support.warning"));
-                    }
-                } catch (MatlabVersionNotFoundException e) {
-                    return FormValidation
-                            .warning(Message.getValue("Builder.invalid.matlab.root.warning"));
-                }
-            }
-            return FormValidation.ok();
-        };
-
-        // Method to get the MatlabRoot value from Build wrapper class.
-        public static String getMatlabRoot() {
-            try {
-                return Jenkins.getInstance().getDescriptorByType(UseMatlabVersionDescriptor.class)
-                        .getMatlabRootFolder();
-            } catch (Exception e) {
-                // For any exception during getMatlabRootFolder() operation, return matlabRoot as
-                // NULL.
-                return null;
-            }
         }
     }
 
@@ -425,8 +298,8 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
      * 7Csort:date/jenkinsci-dev/AFYHSG3NUEI/UsVJIKoE4B8J
      * 
      */
-    public class PdfChkBx {
-        String pdfReportFilePath;
+    public static class PdfChkBx {
+        private String pdfReportFilePath;
 
         @DataBoundConstructor
         public PdfChkBx() {
@@ -443,8 +316,8 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
         }
     }
 
-    public class TapChkBx {
-        String tapReportFilePath;
+    public static class TapChkBx {
+        private String tapReportFilePath;
 
         @DataBoundConstructor
         public TapChkBx() {
@@ -461,8 +334,8 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
         }
     }
 
-    public class JunitChkBx {
-        String junitReportFilePath;
+    public static class JunitChkBx {
+        private String junitReportFilePath;
 
         @DataBoundConstructor
         public JunitChkBx() {
@@ -479,8 +352,8 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
         }
     }
 
-    public class CoberturaChkBx {
-        String coberturaReportFilePath;
+    public static class CoberturaChkBx {
+        private String coberturaReportFilePath;
 
         @DataBoundConstructor
         public CoberturaChkBx() {
@@ -497,8 +370,8 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
         }
     }
 
-    public class StmResultsChkBx {
-        String stmResultsFilePath;
+    public static class StmResultsChkBx {
+        private String stmResultsFilePath;
 
         @DataBoundConstructor
         public StmResultsChkBx() {
@@ -515,8 +388,8 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
         }
     }
 
-    public class ModelCovChkBx {
-        String modelCoverageFilePath;
+    public static class ModelCovChkBx {
+        private String modelCoverageFilePath;
 
         @DataBoundConstructor
         public ModelCovChkBx() {
--- a/src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
+++ b/src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
@@ -43,11 +43,18 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
     private int buildResult;
     private EnvVars env;
     private boolean tapChkBx;
+    private String tapChkBxFilePath;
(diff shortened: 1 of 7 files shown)


src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/config.jelly
```diff
@@ -2,30 +2,43 @@
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
     
 <f:section title="Generate Test Artifacts">
-  	 <f:entry field="pdfReportChkBx">	
-    	<f:checkbox title="PDF test report " name="pdfReportChkBx" checked="${instance.pdfReportChkBx}"/>	
+    <f:optionalBlock name="pdfReportChkBx" field="pdfReportChkBx" title="PDF test report" checked="${instance.isPdfChecked}">
+  	 <f:entry field="pdfReportFilePath" title="File location ">
```
mw-hrastega, line 6: "File location" is fine. "File path" is another possibility.

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/runMatlabTests.m
```diff
@@ -2,22 +2,24 @@
...
-p.addParameter('JUnitResults', false, @islogical);
-p.addParameter('SimulinkTestResults', false, @islogical);
-p.addParameter('CoberturaCodeCoverage', false, @islogical);
-p.addParameter('CoberturaModelCoverage', false, @islogical);
+p.addParameter('PDFReportPath', '', @ischar);
+p.addParameter('TAPResultsPath', '', @ischar);
+p.addParameter('JUnitResultsPath', '', @ischar);
+p.addParameter('SimulinkTestResultsPath', '', @ischar);
+p.addParameter('CoberturaCodeCoveragePath', '', @ischar);
+p.addParameter('CoberturaModelCoveragePath', '', @ischar);
```
mcafaro, line 12: `@ischar` will return true for char matrices. You need to further constrain it so it can only be a char row vector. Something like `@(c)ischar(c) && (isempty(c) || isrow(c))`.

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/config.jelly
nbhoski: Changed to **File path :**

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -316,19 +266,144 @@ public String constructCommandForTest(String inputArguments) {
...
+        }
+
+        return String.join(",", inputArgs);
+    }
+    
+    private void addInputArgs(String reportName, String reportPath) {
+        if (reportPath != null) {
+            inputArgs.add(reportName + "," + "'" + reportPath + "'");
+        }
+    }
```
nbhoski, line 289: @mcafaro  here the code which wraps the input args

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/runMatlabTests.m
nbhoski: Do you see any possibility of char matrices being passed ? . unless used this this file in isolation without our plugin.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -33,86 +20,161 @@
...
+        }
+        return false;
+    }
+      
+    public StmResultsChkBx getStmResultsChkBx() {
+        return this.stmResultsChkBx;
+    } 
+    
+    public String getStmResultsFilePath() {
+        return stmResultsFilePath;
```
mw-akumar, line 145: Just curious, why is there no "this" over here?

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -316,19 +266,144 @@ public String constructCommandForTest(String inputArguments) {
...
+     * Classes for each optional block in jelly file.This is restriction from Stapler architecture
+     * when we use <f:optionalBlock> as it creates a object for each block in JSON. This could be
+     * simplified by using inline=true attribute of <f:optionalBlock> however it has some abrupt UI
+     * scrolling issue on click and also some esthetic issue like broken gray side bar appears.Some
+     * discussion about this on Jenkins forum
+     * https://groups.google.com/forum/#!searchin/jenkinsci-dev/OptionalBlock$20action$20class%
+     * 7Csort:date/jenkinsci-dev/AFYHSG3NUEI/UsVJIKoE4B8J
+     * 
+     */
+    public static class PdfChkBx {
```
mw-akumar, line 301: Given the boiler plate nature of this code, this almost calls for a class generator like infrastructure :-). Perhaps not needed now, but something to invest in for the future.

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/runMatlabTests.m
```diff
@@ -32,35 +34,33 @@
 import('matlab.unittest.TestRunner');
 runner = TestRunner.withTextOutput;
 
-% Add the requested plugins
-resultsDir = fullfile(pwd, 'matlabTestArtifacts');
+
 
 % Produce JUnit report
-if produceJUnit
+if ~isempty(junitReportPath)
```
mw-akumar, line 40: The more I look at this, the more I think that it needs refactoring. Again, not something that needs to be done now, but something worth considering.

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/runMatlabTests.m
mcafaro: I tend to think a well design component should work in isolation. Sometimes some cost/benefit analysis to consider but this one seems fairly low cost.

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/runMatlabTests.m
acampbel: Really we should remove this file entirely from resources and leverage the script generator. It should already be extensible, works across all CI platforms we support, and produces a nice simple script with non conditional code that the customer can read and leverage/reuse/modify/etc. Said another way, I think the refactoring has been done already, we just need to leverage it. I think delivering pipelines support is the priority here, but theres a follow up task this release to use the scriptgen infrastructure.

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/runMatlabTests.m
```diff
@@ -133,11 +131,18 @@
 results = runner.run(suite);
 failed = any([results.Failed]);
 
+function fileName = getFullFileForReport(filePath)
+[filepath,name,ext] = fileparts(filePath);
+if (filepath == "")
```
acampbel, line 136: This will not work on versions of MATLAB that don't support strings. I am not sure why this is necessary anyway. It doesn't seem right to me that we deconstruct the file parts, sometimes mkdir if needed, and then put the file back together and return the same string (except when we are in pwd and we then move to an absolute path instead of a relative one, which is inconsistent with other relative paths. Also, it is named getFullfileForReport but it accepts the full file for the report and returns the full file for the report but has side effects.

Rather than doing this, can we simply call mkdirIfNeeded with the file path? Could either be:

`function mkdirIfNeeded(dir)
dirExists = isempty(dir) || exist(dir,'dir') == 7;
if ~dirExists		
    mkdir(dir);		    
end	`

where we call it like so: `mkdirIfNeeded(fileparts(pdfReportPath))`

or we just pass in the whole path each time and we issue the file parts in mkdirIfNeeded:

`function preparePath(path)
dir = fileparts(path);
dirExists = isempty(dir) || exist(dir,'dir') == 7;
if ~dirExists		
    mkdir(dir);		    
end`

Note maybe preparePAth is a better name for it at that point, and that is a name that implies possible side effects (like creating a dir)

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
nbhoski: it was missed :) thanks for pointing it out

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/runMatlabTests.m
nbhoski: Done.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -1,27 +1,14 @@
 package com.mathworks.ci;
 
-/**
- * Copyright 2019-2020 The MathWorks, Inc.
- * 
- * MATLAB test run builder used to run all MATLAB & Simulink tests automatically and generate
- * selected test artifacts.
- * 
- */
```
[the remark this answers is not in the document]
nbhoski: Done.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -33,86 +20,161 @@
...
-    private boolean coberturaChkBx;
-    private boolean stmResultsChkBx;
-    private boolean modelCoverageChkBx;
-    private boolean pdfReportChkBx;
+    private TapChkBx tapChkBx;
+    private JunitChkBx junitChkBx;   
+    private CoberturaChkBx coberturaChkBx;
+    private StmResultsChkBx stmResultsChkBx; 
+    private ModelCovChkBx modelCoverageChkBx; 
+    private PdfChkBx pdfReportChkBx;
```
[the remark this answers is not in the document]
nbhoski: This is needed as @Databountsetters are tightly coupled with the fields on the Jelly . to store and set the value of the the fields in jelly we should have appropriate public getters and setters  for each field in the builder class.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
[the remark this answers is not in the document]
nbhoski: Done

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/runMatlabTests.m
nbhoski: Done. implemented preparePath()

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -33,86 +28,161 @@
...
         this.modelCoverageChkBx = modelCoverageChkBx;
-    }
+        this.modelCoverageFilePath = this.modelCoverageChkBx.getModelCoverageFilePath();
+    }   
 
     @DataBoundSetter
-    public void setPdfReportChkBx(boolean pdfReportChkBx) {
+    public void setPdfReportChkBx(PdfChkBx pdfReportChkBx) {
         this.pdfReportChkBx = pdfReportChkBx;
+        this.pdfReportFilePath = this.pdfReportChkBx.getPdfReportFilePath();
```
acampbel, line 95: Why hold onto two string? Why not simply access the file path directly from the  instance of what you call `pdfReportChkBx`? (I like `pdfArtifact` as a better name for that)

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -33,86 +28,161 @@
...
     }
-
-    public boolean getTapChkBx() {
-        return tapChkBx;
+    
+    public String getTapReportFilePath() {
+        return this.tapReportFilePath;
+    }      
+    public boolean getIsPdfChecked() {
```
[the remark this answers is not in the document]
acampbel: Nevermind I see the way they are being invoked from the jelly script may require this. Does the jelly script work if you just name isFoo and reference it as foo from the jelly?

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/config.jelly
```diff
@@ -2,30 +2,43 @@
...
+    <f:optionalBlock name="pdfReportChkBx" field="pdfReportChkBx" title="PDF test report" checked="${instance.isPdfChecked}">
+  	 <f:entry field="pdfReportFilePath" title="File path: ">	
+    	<f:textbox default="matlabTestArtifacts/testreport.pdf"/>
      </f:entry>
-
-	 <f:entry field="tapChkBx">
-    	<f:checkbox title="TAP test results" name="tapChkBx" checked="${instance.tapChkBx}"/>
+    </f:optionalBlock>
+     
+    <f:optionalBlock name="tapChkBx" field="tapChkBx" title="TAP test results" checked="${instance.isTapChecked}">
```
acampbel, line 11: I don't see how/why the jelly "queries" the isTapChecked method. It seems like the selections in the UI should set the artifacts, but I don't see why the jelly would need to query them.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -316,19 +274,144 @@ public String constructCommandForTest(String inputArguments) {
...
-
-        return inputArgsToMatlabFcn;
+        addInputArgs(MatlabBuilderConstants.PDF_REPORT_PATH, getPdfReportFilePath());
+        addInputArgs(MatlabBuilderConstants.TAP_RESULTS_PATH, getTapReportFilePath());
+        addInputArgs(MatlabBuilderConstants.JUNIT_RESULTS_PATH, getJunitReportFilePath());
+        addInputArgs(MatlabBuilderConstants.STM_RESULTS_PATH, getStmResultsFilePath());
+        addInputArgs(MatlabBuilderConstants.COBERTURA_CODE_COVERAGE_PATH,
+                getCoberturaReportFilePath());
+        addInputArgs(MatlabBuilderConstants.COBERTURA_MODEL_COVERAGE_PATH,
+                getModelCoverageFilePath());
```
acampbel, line 284: To me this is pointing to some logic that needs to be contained in other classes (not the builder). For example, if the various artifact classes that you link to the checkbox fields had an `appendInputArguments` method, or a `contributeTo(ScratchInputArguments args)` then each of the artifact classes could actually do this work, and the bud,er wouldn't need to. Also, this would allow moving all these constants to the actual artifact classes you have created, which seem like the right location for that knowledge anyway.

Finally, this would allow you to leverage the NullObject pattern so that you didn't need any null checks. Instead, the default value for the artifacts would be a NullArtifact which would have a contributeTo (or whatever its called) method that simply does nothing.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -33,86 +28,161 @@
...
-        return tapChkBx;
+    
+    public String getTapReportFilePath() {
+        return this.tapReportFilePath;
+    }      
+    public boolean getIsPdfChecked() {
+        if(this.pdfReportChkBx != null) {
+            return true;
+        }
+        return false;
```
acampbel, line 105: Does null always mean false? I am thinking of the case where the check box was checked, the. alter unchecked. Is this stateful such that it is unchecked and thus should be false? Worried about bugs here with this logic. If there is no bug, this can be simplified to:

    return this.pdfReportChkBx != null

...but really I prefer the approach of using the NullObject pattern and avoiding the raw nulls entirely. See a later comment.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
acampbel: or better yet, the builder never even needs to know about these locations at all! If the artifact classes are the things that contribute to the inputargs then the builder never needs this knowledge at all

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
acampbel, line 101: Can you put this method next to the other pdf related methods?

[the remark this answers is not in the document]
nbhoski: No  it does not work it always searches for get and then the attribute name with First letter in upper case.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -316,19 +274,144 @@ public String constructCommandForTest(String inputArguments) {
...
+    public static class TapChkBx {
+        private String tapReportFilePath;
+
+        @DataBoundConstructor
+        public TapChkBx() {
+
+        }
+
+        @DataBoundSetter
+        public void setTapReportFilePath(String tapReportFilePath) {
```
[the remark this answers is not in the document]
nbhoski: Actually in code we never do the set() operation. Again this brings us to gelly constraints where each field value name should be unique. if i do setFilePath the i will have to change each text field names in jelly as filePath and we cant have same field name for each jelly text box.  I tried setting up the same name however it did not work.

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/config.jelly
nbhoski: Usually jelly queries the values to retain the previous selected values ans reflect it on UI when User revisits the configuration page after performing save operation on page. hence to show whether this option was checked or not we had to call this. However i have replaced it with inline expression in jelly that actually checks for the object for null and set the value accordingly.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
nbhoski: Yes, NULL here always means false. I have already removed all these methods i could achive the same in jelly by inline expressions instead.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
nbhoski: Done. its now removed anyways.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -316,19 +274,144 @@ public String constructCommandForTest(String inputArguments) {
...
+        public void setPdfReportFilePath(String pdfReportFilePath) {
+            this.pdfReportFilePath = pdfReportFilePath;
+        }
+
+        public String getPdfReportFilePath() {
+            return this.pdfReportFilePath;
+        }
+    }
+
+    public static class TapChkBx {
```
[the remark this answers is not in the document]
nbhoski: Done. changed class names as Artifact class

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
nbhoski: names changed as per your suggestion.

nbhoski: using of two identical fields seems to be design constraints if we use optional block with inline=false (which we did because of jittery UI) I tried isolating these fields to keep it only in the scope of the specific classes however to resurrect the values back on UI and also to apply the default values to the field maintaining these in builder class seems to be required. i did not find any official gude on how to use optional block however this is the reference where it is explained on how optional blocks could be used. https://stackoverflow.com/questions/13024929/how-to-use-optionalblock-in-build-steps-config-jelly

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
nbhoski: Added addFilePathTo() method for each artifact class and moved all the constants to the specific artifact classes. However i am facing few issues to leverage the same for NullObject pattern. the major problem is the checkboxes are cheked on UI when its not NULL and in the null patteern if nothing is set it still NullArtifact by default hence on UI it appears all the values as checked I might need to discuss with you or team to see how i could handle this using NullObject pattern . meanwhile i hav eupdated the code for isolating the  addFilePathTo() method to each artifact class however here i again will hav eto check for the NULL object anyways.

src/test/java/com/mathworks/ci/RunMatlabTestsBuilderTester.java
```diff
@@ -27,12 +27,12 @@
...
-    private boolean coberturaChkBx;
-    private boolean stmResultsChkBx;
-    private boolean modelCoverageChkBx;
-    private boolean pdfReportChkBx;
+    private TapArtifact tapChkBx;
+    private JunitArtifact junitChkBx;
+    private CoberturaArtifact coberturaChkBx;
+    private StmResultsArtifact stmResultsChkBx;
+    private ModelCovArtifact modelCoverageChkBx;
+    private PdfArtifact pdfReportChkBx;
```
mcafaro, line 35: Seems like these variable names need to be updated to reflect the changes you've made to their type?

PR #88, opened 2020-05-11
Artifact input change nop

nbhoski:
Hi Andy , tried your suggestion and figured out the following things.

* Need to maintain the duplicate strings in builder class else it would not resurrect on screen .
* we cant use Interface and / or Abstract class to implement null object pattern  as stapler  class always try to instantiate the class that it has been made a type of 

eg : Artifcat tapArtifact = new NullArtifact();

Stapler will always try to instantiate  Artifcat and it fails when its either abstract or interface with error like below.

Caused by: org.kohsuke.stapler.NoStaplerConstructorException: There's no @DataBoundConstructor on any constructor of interface com.mathworks.ci.RunMatlabTestsBuilder$Artifact
	at org.kohsuke.stapler.ClassDescriptor.loadConstructorParamNames(ClassDescriptor.java:247)
	at org.kohsuke.stapler.RequestImpl.instantiate(RequestImpl.java:758)
	at org.kohsuke.stapler.RequestImpl.access$200(RequestImpl.java:83)
	at org.kohsuke.stapler.RequestImpl$TypePair.convertJSON(RequestImpl.java:674)

* in this draft i have tried to use a Artifact as parent class and tried to implement the null object pattern however i am kind of fuguring out on how i could handdle the UI checkbox issue without checking for NULL in the builder class.

### DIFF
--- a/src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
+++ b/src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
@@ -1,27 +1,14 @@
 package com.mathworks.ci;
 
-/**
- * Copyright 2019-2020 The MathWorks, Inc.
- * 
- * MATLAB test run builder used to run all MATLAB & Simulink tests automatically and generate
- * selected test artifacts.
- * 
- */
-
-
-import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.function.Function;
 import javax.annotation.Nonnull;
 import org.apache.commons.io.FilenameUtils;
 import org.jenkinsci.Symbol;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.DataBoundSetter;
-import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.StaplerRequest;
-import com.mathworks.ci.UseMatlabVersionBuildWrapper.UseMatlabVersionDescriptor;
 import hudson.EnvVars;
 import hudson.Extension;
 import hudson.FilePath;
@@ -33,8 +20,6 @@ import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.tasks.BuildStepDescriptor;
 import hudson.tasks.Builder;
-import hudson.util.FormValidation;
-import jenkins.model.Jenkins;
 import jenkins.tasks.SimpleBuildStep;
 import net.sf.json.JSONObject;
 
@@ -202,8 +187,6 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
     @Extension
     public static class RunMatlabTestsDescriptor extends BuildStepDescriptor<Builder> {
 
-        MatlabReleaseInfo rel;
-
         // Overridden Method used to show the text under build dropdown
         @Override
         public String getDisplayName() {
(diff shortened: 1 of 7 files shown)


nbhoski:
Thanks @mw-akumar  for your inputs. I shall do the necessary changes and add that to final review 
@acampbel  this seems to work i shall make some necessary changes and shall send for final review.

nbhoski:
> Hi Andy , tried your suggestion and figured out the following things.
> 
> * Need to maintain the duplicate strings in builder class else it would not resurrect on screen .
> * we cant use Interface and / or Abstract class to implement null object pattern  as stapler  class always try to instantiate the class that it has been made a type of
> 
> eg : Artifcat tapArtifact = new NullArtifact();
> 
> Stapler will always try to instantiate Artifcat and it fails when its either abstract or interface with error like below.
> 
> Caused by: org.kohsuke.stapler.NoStaplerConstructorException: There's no @DataBoundConstructor on any constructor of interface com.mathworks.ci.RunMatlabTestsBuilder$Artifact
> at org.kohsuke.stapler.ClassDescriptor.loadConstructorParamNames(ClassDescriptor.java:247)
> at org.kohsuke.stapler.RequestImpl.instantiate(RequestImpl.java:758)
> at org.kohsuke.stapler.RequestImpl.access$200(RequestImpl.java:83)
> at org.kohsuke.stapler.RequestImpl$TypePair.convertJSON(RequestImpl.java:674)
> 
> * in this draft i have tried to use a Artifact as parent class and tried to implement the null object pattern however i am kind of fuguring out on how i could handdle the UI checkbox issue without checking for NULL in the builder class.

nbhoski:
Sorry My bad. I have implemented the changes and re opened the PR

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -491,29 +442,16 @@ public boolean getDefault() {
...
-        public String getFilePath() {
-            return null;
-        }
-        
+    public static abstract class Artifact {
+        public abstract void addFilePathArgTo(List<String> inputArgs);
+
+        public abstract String getFilePath();
+
         public boolean getDefault() {
```
mw-akumar, line 453: I think the name getDefault should be revisited. Perhaps getNullability? (sorry I couldn't seem to think of a better name :-) )

nbhoski: is it good idea to use isSelected() or selected() ?  i know its more aligned to functional behavior of UI. but i am running short of names either :)

mcafaro:
I'm confused. Why does this pull request appear to contain all of the changes from pull request #86 ?

acampbel:
> I'm confused. Why does this pull request appear to contain all of the changes from pull request #86 ?

acampbel:
> I'm confused. Why does this pull request appear to contain all of the changes from pull request #86 ?

I think this can be considered part of the other code review, I believe he made it on another branhc. I think there is even another one that superseded this one as well, where he implemented the NullObject pattern and cleaned up the artifact handling implementation a bit. I have approved that one, kind of as an approval of where we have gotten to our all of them.

nbhoski:
All Final changes are here now, Earlier PR were branched out of this PR as those were experimental . it would have been easier to fall back on the working state if our experimental branches would not have worked. All our approved changes are here in this PR now i will merge this once the azure pipeline passes. and close all other PR's

src/test/java/com/mathworks/ci/RunMatlabTestsBuilderTester.java
```diff
@@ -27,12 +27,12 @@
...
-    private boolean coberturaChkBx;
-    private boolean stmResultsChkBx;
-    private boolean modelCoverageChkBx;
-    private boolean pdfReportChkBx;
+    private TapArtifact tapChkBx;
+    private JunitArtifact junitChkBx;
+    private CoberturaArtifact coberturaChkBx;
+    private StmResultsArtifact stmResultsChkBx;
+    private ModelCovArtifact modelCoverageChkBx;
+    private PdfArtifact pdfReportChkBx;
```
mcafaro, line 35: Think these variable names need to be updated.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -316,19 +224,171 @@ public String constructCommandForTest(String inputArguments) {
...
+
+    public static class NullArtifact implements Artifact {
+
+        @Override
+        public void addFilePathArgTo(Map<String, String> inputArgs) {
+
+        }
+
+        @Override
+        public boolean getDefault() {
```
acampbel, line 354: Since this is returning false for the unchecked case, you really should rename this to be getSelected or getIsSelected, and the resulting jelly file of course would need updating

acampbel: (or name it something like that. getDefault is wrong now because default applies to when the artifact is not selected so should return true for that naming)

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/runMatlabTests.m
```diff
@@ -133,13 +132,18 @@
...
 failed = any([results.Failed]);
 
+function preparePath(path)
+dir = fileparts(path);
+dirExists = isempty(dir) || exist(dir,'dir') == 7;
+if ~dirExists		
+    mkdir(dir);		    
+end
 
 function tapToFile = getTapResultFile(resultsDir)
```
acampbel, line 142: This variable should be called tapFile

acampbel, line 142: This variable should be called something like tapFile.

src/test/java/com/mathworks/ci/RunMatlabTestsBuilderTester.java
nbhoski: Done .

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
nbhoski: Done.

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/runMatlabTests.m
nbhoski: Done

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/runMatlabTests.m
```diff
@@ -133,13 +132,18 @@
...
 
+function preparePath(path)
+dir = fileparts(path);
+dirExists = isempty(dir) || exist(dir,'dir') == 7;
+if ~dirExists		
+    mkdir(dir);		    
+end
 
-function tapToFile = getTapResultFile(resultsDir)
+function tapFile = getTapResultFile(resultsDir)
```
acampbel, line 142: Funny, I was referring to the other variable, "resultsDir" to change. This is not a dir it is the fullpath to the desired tap file. Thats why it is confusing, you can't fopen a dir for example, so the code is misleading.

This can wait to change in a follow-up commit.

PR #90, opened 2020-05-14
Initial Changes to use data bound constructor.

nbhoski:
* Used @databound constructors for each artifact class.
* Reduced scope of constant valuables to Artifact specific class.
* Single quote logic is scoped to Inputargs() method of builder class.

### DIFF
--- a/src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
+++ b/src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
@@ -11,8 +11,11 @@ package com.mathworks.ci;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import javax.annotation.Nonnull;
+import org.apache.commons.collections.map.HashedMap;
 import org.apache.commons.io.FilenameUtils;
 import org.jenkinsci.Symbol;
 import org.kohsuke.stapler.DataBoundConstructor;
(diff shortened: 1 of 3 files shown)


src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -247,149 +254,141 @@ private String getInputArguments() {
...
         @Override
         public boolean getDefault() {
-            
             return false;
         }
 
         @Override
         public String getFilePath() {
-            
             return null;
```
acampbel, line 360: Suggest returning "" instead of null

nbhoski: This has to be NULL due to default value in config.jelly  file . UI shows the default value only if the getter of that field is NULL. as we will assign Nullartifact by default after first save operation default values will not be populated on UI as NullArtifact returns  empty string.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -246,212 +253,142 @@ private String getInputArguments() {
...
      */
-    public static class PdfArtifact extends Artifact {
-        private String pdfReportFilePath;
+    public static class PdfArtifact extends AbstractArtifactImpl {
 
-        @DataBoundConstructor
-        public PdfArtifact() {
-           
-        }
+        static final String PDF_REPORT_PATH = "PDFReportPath";
```
acampbel, line 258: why not make all these private static final?

nbhoski: Done.

Issue #92, opened 2020-05-20
Missing help icon for Test Artifacts with the 'Text Input' option


SangaviJayagopi:

SangaviJayagopi:
Verified for help icon.

Verified for backward compatibility

PR #98, opened 2020-05-22
Supporting backward compatibillity

nbhoski:
supporting backward compatibility for newly  added artifacts.

### DIFF
--- a/src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
+++ b/src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
@@ -39,6 +39,15 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
 
     private int buildResult;
     private EnvVars env;
+    
+    // Make all old values transient which protects them writing back on disk.
+    private transient boolean tapChkBx;
+    private transient boolean junitChkBx;
+    private transient boolean coberturaChkBx;
+    private transient boolean stmResultsChkBx;
+    private transient boolean modelCoverageChkBx;
+    private transient boolean pdfReportChkBx;
+    
     private Artifact tapArtifact = new NullArtifact();
     private Artifact junitArtifact = new NullArtifact();
     private Artifact coberturaArtifact = new NullArtifact();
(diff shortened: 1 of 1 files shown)


src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -140,7 +149,42 @@ private void setEnv(EnvVars env) {
...
+            this.coberturaArtifact = new CoberturaArtifact("matlabTestArtifacts/cobertura.xml");
+        }
+        if (stmResultsChkBx) {
+            this.stmResultsArtifact =
+                    new StmResultsArtifact("matlabTestArtifacts/simulinktestresults.mldatx");
+        }
+        if (modelCoverageChkBx) {
+            this.pdfReportArtifact =
+                    new ModelCovArtifact("matlabTestArtifacts/coberturamodelcoverage.pdf");
+        }
```
mcafaro, line 183: Why is this creating a `pdfReportArtifact`?

nbhoski: Nice Catch :P  Updated it

mcafaro: Shouldn't it have `.xml` extension as well?

nbhoski: Yep :) done 1

mcafaro: Also, why is this a different value than the default in jelly? One is `coberturamodelcoverage.xml` and the other is `modelcoverage.xml`. Seems like these should be the same.

https://github.com/mathworks/jenkins-matlab-plugin/blob/a8f6ce39a9582e6dc887c7f69b9dc20220a254af/src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/config.jelly#L38-L42

nbhoski: This change is already merged in last PR #92

mcafaro: Which one is right?

nbhoski: The One which is merged is correct . That why i did not include that file in this PR   	
below is correct and already merged in the qualification branch 
  <f:textbox default="matlabTestArtifacts/coberturamodelcoverage.xml"/>

nbhoski: https://github.com/mathworks/jenkins-matlab-plugin/blob/6f5ac8c8a322a8b96ae1d1e1df65178e18c2ac3c/src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/config.jelly#L40

PR #112, opened 2020-06-18
Matlab cmd as script

nbhoski:
* Added support to run MATLAB command through script file to address the issues of "" and other special characters.
* Writing the command in command_UUID.m script file and passing that as a parameter to run_matalb_command(bat|sh) script .
* cleaning the entire temp folder after job completion.
* Few tests which were based on console output of the command has been disabled as now commands cannot be seen on the console window as it will be wrapped in temp script file . I shall add them back to integ-tests once integ-test folders merged.

### DIFF
--- a/src/main/java/com/mathworks/ci/MatlabBuild.java
+++ b/src/main/java/com/mathworks/ci/MatlabBuild.java
@@ -44,7 +44,7 @@ public interface MatlabBuild {
                     targetWorkspace);
         } else {
             final String runnerScriptName = uniqueName + "\\run_matlab_command.bat";
-            matlabLauncher.cmds(tmpDir + "\\" + runnerScriptName, "\"" + matlabCommand + "\"")
+            matlabLauncher.cmds("cmd.exe","/C",tmpDir + "\\" + runnerScriptName, "\"" + matlabCommand + "\"")
                     .stdout(listener);
             // Copy runner.bat for Windows platform in workspace.
             copyFileInWorkspace(MatlabBuilderConstants.BAT_RUNNER_SCRIPT, runnerScriptName,
@@ -86,13 +86,5 @@ public interface MatlabBuild {
 
     default String getUniqueNameForRunnerFile() {
         return UUID.randomUUID().toString();
-    }
-    
-    default Launcher getDecoratedLauncherForWindows(Launcher launcher) {
-      //decorate the launcher if its on windows 
-        if(!launcher.isUnix()) {
-            launcher = launcher.decorateByPrefix("cmd.exe", "/C");
-        }
-        return launcher;
     }
 }
--- a/src/main/java/com/mathworks/ci/MatlabBuild.java
+++ b/src/main/java/com/mathworks/ci/MatlabBuild.java
@@ -22,33 +22,29 @@ public interface MatlabBuild {
      * This Method decorates the launcher with MATLAB command provided and returns the Process
      * object to launch MATLAB with appropriate startup options like -r or -batch
      * 
+     * @param matlabLauncher current build launcher
      * @param workspace Current build workspace
      * @param launcher Current build launcher
      * @param listener Current build listener
-     * @param envVars Environment variables of the current build
      * @param matlabCommand MATLAB command to execute on shell
      * @return matlabLauncher returns the process launcher to run MATLAB commands
      */
-    default ProcStarter getProcessToRunMatlabCommand(FilePath workspace, Launcher launcher,
-            TaskListener listener, EnvVars envVars, String matlabCommand, String uniqueName)
+    default ProcStarter getProcessToRunMatlabCommand(ProcStarter matlabLauncher, FilePath workspace,
+            Launcher launcher, TaskListener listener, String matlabCommand, String uniqueName)
             throws IOException, InterruptedException {
         // Get node specific tmp directory to copy matlab runner script
         String tmpDir = getNodeSpecificTmpFolderPath(workspace);
         FilePath targetWorkspace = new FilePath(launcher.getChannel(), tmpDir);
-        ProcStarter matlabLauncher;
         if (launcher.isUnix()) {
             final String runnerScriptName = uniqueName + "/run_matlab_command.sh";
-            matlabLauncher = launcher.launch().pwd(workspace).envs(envVars)
-                    .cmds(tmpDir + "/" + runnerScriptName, matlabCommand).stdout(listener);
+            matlabLauncher.cmds(tmpDir + "/" + runnerScriptName, matlabCommand).stdout(listener);
 
             // Copy runner .sh for linux platform in workspace.
             copyFileInWorkspace(MatlabBuilderConstants.SHELL_RUNNER_SCRIPT, runnerScriptName,
                     targetWorkspace);
         } else {
             final String runnerScriptName = uniqueName + "\\run_matlab_command.bat";
-            launcher = launcher.decorateByPrefix("cmd.exe", "/C");
-            matlabLauncher = launcher.launch().pwd(workspace).envs(envVars)
-                    .cmds(tmpDir + "\\" + runnerScriptName, "\"" + matlabCommand + "\"")
+            matlabLauncher.cmds(tmpDir + "\\" + runnerScriptName, "\"" + matlabCommand + "\"")
                     .stdout(listener);
             // Copy runner.bat for Windows platform in workspace.
             copyFileInWorkspace(MatlabBuilderConstants.BAT_RUNNER_SCRIPT, runnerScriptName,
@@ -90,5 +86,13 @@ public interface MatlabBuild {
 
     default String getUniqueNameForRunnerFile() {
         return UUID.randomUUID().toString();
+    }
+    
+    default Launcher getDecoratedLauncherForWindows(Launcher launcher) {
+      //decorate the launcher if its on windows 
+        if(!launcher.isUnix()) {
+            launcher = launcher.decorateByPrefix("cmd.exe", "/C");
+        }
+        return launcher;
     }
 }
--- a/src/main/java/com/mathworks/ci/MatlabBuild.java
+++ b/src/main/java/com/mathworks/ci/MatlabBuild.java
@@ -22,21 +22,23 @@ public interface MatlabBuild {
      * This Method decorates the launcher with MATLAB command provided and returns the Process
      * object to launch MATLAB with appropriate startup options like -r or -batch
      * 
-     * @param matlabLauncher current build launcher
      * @param workspace Current build workspace
      * @param launcher Current build launcher
      * @param listener Current build listener
+     * @param envVars Environment variables of the current build
      * @param matlabCommand MATLAB command to execute on shell
      * @return matlabLauncher returns the process launcher to run MATLAB commands
      */
-    default ProcStarter getProcessToRunMatlabCommand(ProcStarter matlabLauncher, FilePath workspace,
-            Launcher launcher, TaskListener listener, String matlabCommand, String uniqueName)
+    default ProcStarter getProcessToRunMatlabCommand(FilePath workspace,
+            Launcher launcher, TaskListener listener, EnvVars envVars, String matlabCommand, String uniqueName)
             throws IOException, InterruptedException {
         // Get node specific tmp directory to copy matlab runner script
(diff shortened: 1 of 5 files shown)


src/main/java/com/mathworks/ci/MatlabBuild.java
```diff
@@ -22,33 +22,29 @@
...
             throws IOException, InterruptedException {
         // Get node specific tmp directory to copy matlab runner script
         String tmpDir = getNodeSpecificTmpFolderPath(workspace);
         FilePath targetWorkspace = new FilePath(launcher.getChannel(), tmpDir);
-        ProcStarter matlabLauncher;
         if (launcher.isUnix()) {
             final String runnerScriptName = uniqueName + "/run_matlab_command.sh";
-            matlabLauncher = launcher.launch().pwd(workspace).envs(envVars)
-                    .cmds(tmpDir + "/" + runnerScriptName, matlabCommand).stdout(listener);
+            matlabLauncher.cmds(tmpDir + "/" + runnerScriptName, matlabCommand).stdout(listener);
```
mcafaro, line 40: Why does the `runnerScriptName` start with a forward slash and then `tmpDir` is separated from it with a forward slash as well?

nbhoski: That's because I create a tmpDir as UUID in keep it in system tmp directory with all helper files like runner script and now Matlab script and delete the entire folder at the end.

src/main/java/com/mathworks/ci/RunMatlabCommandBuilder.java
```diff
@@ -112,22 +116,40 @@ public void perform(@Nonnull Run<?, ?> build, @Nonnull FilePath workspace,
     private synchronized int execMatlabCommand(FilePath workspace, Launcher launcher,
             TaskListener listener, EnvVars envVars) throws IOException, InterruptedException {
         final String uniqueTmpFldrName = getUniqueNameForRunnerFile();
-        ProcStarter matlabLauncher;
+        final String uniqueCommandFile =
+                "command_" + getUniqueNameForRunnerFile().replaceAll("-", "_");
+
+        // Get unique temporary folder filepath
+        final FilePath uniqeTmpFolderPath =
```
mw-akumar, line 123: At the very least the creation of the temporary command file should be its own function so that we don't have all these implementation details here.

src/main/java/com/mathworks/ci/RunMatlabCommandBuilder.java
```diff
@@ -112,22 +116,40 @@ public void perform(@Nonnull Run<?, ?> build, @Nonnull FilePath workspace,
...
+        final FilePath matlabCommandFile =
+                new FilePath(uniqeTmpFolderPath, uniqueCommandFile + ".m");
+        final String matlabCommandFileContent =
+                "cd '" + workspace.getRemote().replaceAll("'", "''") + "';\n" + getCommand();
+
+        // Display the commands on console output for users reference
+        listener.getLogger()
+                .println("Generating MATLAB script with content:\n" + getCommand() + "\n");
+
+        matlabCommandFile.write(matlabCommandFileContent, "UTF-8");
```
mw-akumar, line 135: MATLAB supports UTF-8 since 20a. Versions prior to 20a may not work correctly if we have non-ASCII characters in the command. We may have to document limitations regarding non-ASCII characters.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -238,14 +238,18 @@ private synchronized int execMatlabCommand(FilePath workspace, Launcher launcher
         final String uniqueTmpFldrName = getUniqueNameForRunnerFile();
         ProcStarter matlabLauncher;
         try {
-            matlabLauncher = getProcessToRunMatlabCommand(workspace, launcher, listener, envVars,
-                    constructCommandForTest(getInputArguments()), uniqueTmpFldrName);
+            // Get decorated launcher for Windows
+            launcher = getDecoratedLauncherForWindows(launcher);
```
mw-akumar, line 242: The decoration of cmd for Windows was earlier encapsulated by the code that created the ProcStarter. Now it has leaked into the execMATLABCommand. Why is this needed?

src/main/java/com/mathworks/ci/RunMatlabCommandBuilder.java
nbhoski: Done. file creation is moved to separate method.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
nbhoski: Removed decoration usage as it was coming from Launcher now handling it in cmds() method now .

src/main/java/com/mathworks/ci/MatlabBuild.java
```diff
@@ -22,33 +22,29 @@
...
      * @param workspace Current build workspace
      * @param launcher Current build launcher
      * @param listener Current build listener
-     * @param envVars Environment variables of the current build
      * @param matlabCommand MATLAB command to execute on shell
      * @return matlabLauncher returns the process launcher to run MATLAB commands
      */
-    default ProcStarter getProcessToRunMatlabCommand(FilePath workspace, Launcher launcher,
-            TaskListener listener, EnvVars envVars, String matlabCommand, String uniqueName)
+    default ProcStarter getProcessToRunMatlabCommand(ProcStarter matlabLauncher, FilePath workspace,
```
mw-akumar, line 32: This interface change for getProcessToRunMatlabCommand still seems a bit odd and unnecessary. The callers are creating the ProcStarter object, making a change to the cwd, and then calling getProcessToRunMatlabCommand. Instead can we
1. pass the cwd to getProcessToRunMatlabCommand as a parameter. This way the construction of the ProcStarter method is still self contained in one method. 
or
2. set the cwd on the ProcStarter object after calling getProcessToRunMatlabCommand. This way, there is no need to change the interface.

src/main/java/com/mathworks/ci/RunMatlabCommandBuilder.java
```diff
@@ -112,22 +116,45 @@ public void perform(@Nonnull Run<?, ?> build, @Nonnull FilePath workspace,
...
+        final FilePath matlabCommandFile =
+                new FilePath(uniqeTmpFolderPath, uniqueScriptName + ".m");
+        final String matlabCommandFileContent =
+                "cd '" + workspace.getRemote().replaceAll("'", "''") + "';\n" + getCommand();
+
+        // Display the commands on console output for users reference
+        listener.getLogger()
+                .println("Generating MATLAB script with content:\n" + getCommand() + "\n");
+
+        matlabCommandFile.write(matlabCommandFileContent, "UTF-8");
```
mw-akumar, line 158: There is still the open question of UTF-8 support for MATLAB in older releases. I am okay if we just document the limitation.

nbhoski: I will discuss with @mw-hrastega and make sure if its get documented at the least

src/main/java/com/mathworks/ci/MatlabBuild.java
nbhoski: Implemented option 2  its much feasible than option 1 . & yes i could avoid many changes with this.

src/main/java/com/mathworks/ci/RunMatlabCommandBuilder.java
acampbel: Will a UTF-8 encoded file with ascii characters still work in older releases? Or are we just talking abut the case where someone includes a non-ascii character in the command? If its the latter then I  am fine with the limitation. If its the former that is a blocking issue. If we are writing ascii characters to the file as UTF-8 will it work in older releases? If not we should encode it differently

mcafaro: My understanding is Unicode if fully backward-compatible with ASCII (ASCII is a subset of Unicode). So UTF8 files should work fine in older releases as long as the script contains only ASCII characters.

If a user enters a command with a non-ASCII character, it will not work correctly in older releases. But it will not work correctly in older releases even if we write the file as ASCII.

acampbel: I am good writing it as unicode as long as it is qualified against at least one release where MATLAB didn't support it for m-files.

acampbel: it will be a great feature for new releases to have a Japanese hello world for example

mw-akumar: I demonstrated the implications to Nikhil today. First of all, if the users are using ASCII text only then all this is moot. All our supported encodings use the same binary representation for ASCII. 

The problem occurs if the users use non-ASCII characters. For instance a MATLAB user in Japan who uses Japanese characters in their command. e.g. 'disp("こんにちは"). If the user in Japan is using a Japanese locale Windows machine (a very likely case), then MATLAB prior to R2020a will expect the .m files to be encoded the SHIFT-JIS encoding. Executing the UTF-8 encoded file will produce character corruption. In MATLAB 2020a and newer, the utf-8 encoded file with Japanese characters will be executed correctly.
However, this is not a trivial problem to solve for 19b and older, as we will have to ask MATLAB what the right .m encoding (Shift-JIS on a Japanese Windows,  ISO-8859-1 on an English Windows and so on) will be. For this we will have to actually run MATLAB and execute some commands. :-)

nbhoski: Updated the help doc for custom MATLAB command 

image

src/main/resources/com/mathworks/ci/RunMatlabCommandBuilder/help-matlabCommand.html
```diff
@@ -4,7 +4,10 @@
  <b>Run commands:</b> results = runtests(‘IncludingSubfolders’,true); assert(all(~[results.Failed]))<br>
  <b>Run a script:</b> runMyScript<br>
  <br>&nbsp;</br>
- <b>Recommendation:</b>If you require a number of MATLAB commands to execute your build, consider writing a MATLAB script and executing the script file instead.<br>
+ <b>Recommendation:</b><ul><li>If you require a number of MATLAB commands to execute your build, consider writing a MATLAB script and executing the script file instead.<br>
+ </li>
+ <li>Input text box supports commands only in ASCII formt. For locale specific commands, Write a MATLAB script using the supported chracater encoding and execute the script file.</li>
+ </ul>
  <b>Note:</b> The build will fail if the execution of any MATLAB command causes an error.
```
nbhoski, line 11: @mw-hrastega request you to review these help text changes for locale-specific MATLAB command

mw-hrastega: How about something like this?

The **Command** box supports only ASCII characters. If you use non-ASCII characters in your commands, write your commands to a .m file and invoke the script instead.

mw-hrastega: Can we remove the existing recommendation? I find it kind of redundant now.

nbhoski: Done.

image

mw-hrastega: Quick question. Can we also specify a .mlx file? If so, let's change ".m file" to "plain text file".

nbhoski: I think we could run both , However instead of plain text we could use .m/.mlx ?

mw-hrastega: Yes, that would be better. Lets replace ".m" with ".m or .mlx".

nbhoski: changed it to .m or .mlx 


image

PR #119, opened 2020-07-17
Run MATLAB command using cd and invoke MATLAB from workspace.

nbhoski:
* MATLAB is invoked from the workspace instead of invoking it from tmp folder.
* MATLAB command script will be invoked from tmp folder using 'cd' command. ie from workspace first 'cd' to tmp folder and invoke the custom .m script.
* Added tests to verify this behavior

### DIFF
--- a/src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
+++ b/src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
@@ -72,7 +72,7 @@ public class MatlabCommandStepExecution extends StepExecution implements MatlabB
 
         try {
             matlabLauncher = getProcessToRunMatlabCommand(workspace, launcher, listener, envVars,
-                    "cd('"+ uniqeTmpFolderPath.getRemote() +"');"+ uniqueCommandFile, uniqueTmpFldrName);
+                    "cd('"+ uniqeTmpFolderPath.getRemote().replaceAll("'", "''") +"');"+ uniqueCommandFile, uniqueTmpFldrName);
             listener.getLogger()
                     .println("#################### Starting command output ####################");
             return matlabLauncher.pwd(workspace).join();
--- a/src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
+++ b/src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
@@ -72,11 +72,10 @@ public class MatlabCommandStepExecution extends StepExecution implements MatlabB
 
         try {
             matlabLauncher = getProcessToRunMatlabCommand(workspace, launcher, listener, envVars,
-                    uniqueCommandFile, uniqueTmpFldrName);
-            launcher.launch().pwd(uniqeTmpFolderPath).envs(envVars);
+                    "cd('"+ uniqeTmpFolderPath.getRemote() +"');"+ uniqueCommandFile, uniqueTmpFldrName);
             listener.getLogger()
                     .println("#################### Starting command output ####################");
-            return matlabLauncher.pwd(uniqeTmpFolderPath).join();
+            return matlabLauncher.pwd(workspace).join();
 
         } catch (Exception e) {
             listener.getLogger().println(e.getMessage());
--- a/src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
+++ b/src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
@@ -63,16 +63,16 @@ public class MatlabCommandStepExecution extends StepExecution implements MatlabB
         final String uniqueTmpFldrName = getUniqueNameForRunnerFile();
         final String uniqueCommandFile =
                 "command_" + getUniqueNameForRunnerFile().replaceAll("-", "_");
-        final FilePath uniqeTmpFolderPath =
+        final FilePath uniqueTmpFolderPath =
                 getFilePathForUniqueFolder(launcher, uniqueTmpFldrName, workspace);
 
         // Create MATLAB script
-        createMatlabScriptByName(uniqeTmpFolderPath, uniqueCommandFile, workspace, listener);
+        createMatlabScriptByName(uniqueTmpFolderPath, uniqueCommandFile, workspace, listener);
         ProcStarter matlabLauncher;
 
         try {
             matlabLauncher = getProcessToRunMatlabCommand(workspace, launcher, listener, envVars,
-                    "cd('"+ uniqeTmpFolderPath.getRemote().replaceAll("'", "''") +"');"+ uniqueCommandFile, uniqueTmpFldrName);
+                    "cd('"+ uniqueTmpFolderPath.getRemote().replaceAll("'", "''") +"'); "+ uniqueCommandFile, uniqueTmpFldrName);
             listener.getLogger()
                     .println("#################### Starting command output ####################");
             return matlabLauncher.pwd(workspace).join();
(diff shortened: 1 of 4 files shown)


src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
```diff
@@ -72,11 +72,10 @@ private synchronized int execMatlabCommand(FilePath workspace, Launcher launcher
 
         try {
             matlabLauncher = getProcessToRunMatlabCommand(workspace, launcher, listener, envVars,
-                    uniqueCommandFile, uniqueTmpFldrName);
-            launcher.launch().pwd(uniqeTmpFolderPath).envs(envVars);
+                    "cd('"+ uniqeTmpFolderPath.getRemote().replaceAll("'", "''") +"');"+ uniqueCommandFile, uniqueTmpFldrName);
             listener.getLogger()
                     .println("#################### Starting command output ####################");
-            return matlabLauncher.pwd(uniqeTmpFolderPath).join();
+            return matlabLauncher.pwd(workspace).join();
```
mcafaro, line 78: Do you have to set `pwd`? I'm wondering if you don't set it, if it will just properly use the current working folder? Then the user can `cd` in their pipeline and we'll properly launch MATLAB from the folder they are in, rather then assuming they are in the workspace root.

src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
```diff
@@ -72,11 +72,10 @@ private synchronized int execMatlabCommand(FilePath workspace, Launcher launcher
 
         try {
             matlabLauncher = getProcessToRunMatlabCommand(workspace, launcher, listener, envVars,
-                    uniqueCommandFile, uniqueTmpFldrName);
-            launcher.launch().pwd(uniqeTmpFolderPath).envs(envVars);
+                    "cd('"+ uniqeTmpFolderPath.getRemote().replaceAll("'", "''") +"');"+ uniqueCommandFile, uniqueTmpFldrName);
```
mcafaro, line 75: Typo in `unique`. Maybe add a space between the cd(...); and the call to the script?

src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
nbhoski: its not mandatory to set pwd however i have few questions if we did not set it.
* is there any other place than workspaace that user would keep the code ?
* how user would do CD to required directory ? (Using sh command ? )
* even if user sets the directory before invoking MATLAB we will stil invoke the MATLAB from root (by default if we did not set pwd) and the problem of settings.m not getting picked will remain. 
* I tried using cd before our run and it looks like below.

image

src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
nbhoski: Done.

mw-akumar:
I am wondering if we should consider using addpath instead. With 'cd' we have to once CD to the temp folder, and then back to workspace. Using an addpath, and if needed a rmpath at the end, may be a cleaner design.

src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
mcafaro: Looks like the assumption with Jenkins is that each build step resets the pwd to the workspace root, so I think the way you have it is good.

mcafaro:
> I am wondering if we should consider using addpath instead. With 'cd' we have to once CD to the temp folder, and then back to workspace. Using an addpath, and if needed a rmpath at the end, may be a cleaner design.

This has been a point of discussion previously. From my perspective, it doesn't make much of a difference. If we use `addpath`, ideally we would `rmpath` at the top of the temp script so that the global state is not effected during the user's code execution. So we either `cd` there and immediately `cd` back, or we `addpath` the temp script directory and immediately `rmpath` after the script begins executing.

src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
nbhoski: what if we dont use rmpath anyways the addpath remains valid for only perticular session of MATLAB and since the temp folder is unique namespace colision is remote posssibility

acampbel: If we don't rmpath we have a leaky abstraction. If anyone calls "path" they see a random temp folder on the path. Likewise is we cd but don't cd back. The function handle approach enables us to issue a command "pwd" or "path" and get the right answer back

acampbel:
> > I am wondering if we should consider using addpath instead. With 'cd' we have to once CD to the temp folder, and then back to workspace. Using an addpath, and if needed a rmpath at the end, may be a cleaner design.
> 
> This has been a point of discussion previously. From my perspective, it doesn't make much of a difference. If we use `addpath`, ideally we would `rmpath` at the top of the temp script so that the global state is not effected during the user's code execution. So we either `cd` there and immediately `cd` back, or we `addpath` the temp script directory and immediately `rmpath` after the script begins executing.

I initially felt that add path was cleaner too, but now not is sure. They both effect global state and they both require two steps (initial change and cleanup). My first thought was that cd was slower, but I realized I doubt have any evidence of that. I am now leaning toward cd because it is not as close to other visible global state, if that makes sense. In other words calling path will show our new global state next to a bunch of other global state, namely other folders on the path. Using cd only changes one aspect of global state, the current folder, and it is unlinked in any way to anything else other than the implicit behavior of the path that the ciurrent folder is on the path.

My reasoning is tenuous, but so is the other side of the argument (I think). In other words, why is addpath cleaner than cd? Im not sure we have a good answer either way.

mw-akumar:
> > > I am wondering if we should consider using addpath instead. With 'cd' we have to once CD to the temp folder, and then back to workspace. Using an addpath, and if needed a rmpath at the end, may be a cleaner design.
> > 
> > 
> > This has been a point of discussion previously. From my perspective, it doesn't make much of a difference. If we use `addpath`, ideally we would `rmpath` at the top of the temp script so that the global state is not effected during the user's code execution. So we either `cd` there and immediately `cd` back, or we `addpath` the temp script directory and immediately `rmpath` after the script begins executing.
> 
> I initially felt that add path was cleaner too, but now I'm not so sure. They both effect global state and they both require two steps (initial change and cleanup). My first thought was that cd was slower, but I realized I don't have any evidence of that. I am now leaning toward cd because it is not as close to other visible global state, if that makes sense. In other words calling path will show our new global state next to a bunch of other global state, namely other folders on the path. Using cd only changes one aspect of global state, the current folder, and it is unlinked in any way to anything else other than the implicit behavior of the path that the ciurrent folder is on the path.
> 
> My reasoning is tenuous, but so is the other side of the argument (I think). In other words, why is addpath cleaner than cd? Im not sure we have a good answer either way.

I agree, there doesn't seem to be a clear choice here. And most likely it doesn't matter much Perhaps Chirag Gupta or Scott French from the MATLAB Path Manager can team can be consulted. I am quite certain that calling 'cd' in MATLAB changes the process working directory (by calling SetCurrentDirectory on Windows) and hence gave me the impression that this is an heavier operation. However, I don't have hard data to support it either :-).  Perhaps it is the experience in desktop MATLAB, where 'cd', triggers the Current Folder Browser to refresh, which made me recommend 'addpath'. However the CFB refresh should not be an issue in -batch. 
I am fine with the use of 'cd' here.

PR #122, opened 2020-07-29
Fix Java Temp Path discrepancy for Window

abhargav-mw:
Remove extra back slash from temp folder path generation for Windows platform.
Added comment to tests stating Test assumes there's no Matlab installed and no Matlab on Path.

### DIFF
--- a/src/main/java/com/mathworks/ci/MatlabBuild.java
+++ b/src/main/java/com/mathworks/ci/MatlabBuild.java
@@ -45,6 +45,9 @@ public interface MatlabBuild {
             copyFileInWorkspace(MatlabBuilderConstants.SHELL_RUNNER_SCRIPT, runnerScriptName,
                     targetWorkspace);
         } else {
+            if (tmpDir.charAt(tmpDir.length() - 1) == '\\'){
+                tmpDir = tmpDir.substring(0, tmpDir.length() - 1);
+            }
             final String runnerScriptName = uniqueName + "\\run_matlab_command.bat";
             launcher = launcher.decorateByPrefix("cmd.exe", "/C");
             matlabLauncher = launcher.launch().envs(envVars);
@@ -81,11 +84,8 @@ public interface MatlabBuild {
         // Handle Java Temp Path discrepancy for Windows.
         if (!launcher.isUnix()){
             fileSeperator = "\\";
-            System.out.println("In Windows check;");
             if (tmpDir.charAt(tmpDir.length() - 1) == '\\'){
-                System.out.println("tmpdir = " + tmpDir);
                 tmpDir = tmpDir.substring(0, tmpDir.length() - 1);
-                System.out.println("tmpdirAfter = " + tmpDir);
             }
         }
 
--- a/src/main/java/com/mathworks/ci/MatlabBuild.java
+++ b/src/main/java/com/mathworks/ci/MatlabBuild.java
@@ -33,7 +33,7 @@ public interface MatlabBuild {
             Launcher launcher, TaskListener listener, EnvVars envVars, String matlabCommand, String uniqueName)
             throws IOException, InterruptedException {
         // Get node specific tmp directory to copy matlab runner script
-        String tmpDir = getNodeSpecificTmpFolderPath(workspace, launcher);
+        String tmpDir = getNodeSpecificTmpFolderPath(workspace);
         FilePath targetWorkspace = new FilePath(launcher.getChannel(), tmpDir);
         ProcStarter matlabLauncher;
         if (launcher.isUnix()) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/mathworks/ci/MatlabBuild.java
```diff
@@ -75,16 +75,28 @@ default FilePath getFilePathForUniqueFolder(Launcher launcher, String uniqueName
...
         Computer cmp = workspace.toComputer();
         if (cmp == null) {
             throw new IOException(Message.getValue("build.workspace.computer.not.found"));
         }
         String tmpDir = (String) cmp.getSystemProperties().get("java.io.tmpdir");
+
+        // Handle Java temp folder path discrepancy for Windows.
+        if (!launcher.isUnix() && tmpDir.charAt(tmpDir.length() - 1) == '\\'){
+            tmpDir = tmpDir.substring(0, tmpDir.length() - 1);
+        }
```
mcafaro, line 98: Why only handle this case for Windows? I believe `java.io.tmpdir` is configurable and may have a file separator at the end on any system.

src/main/java/com/mathworks/ci/MatlabBuild.java
```diff
@@ -75,16 +75,28 @@ default FilePath getFilePathForUniqueFolder(Launcher launcher, String uniqueName
...
-    default String getNodeSpecificTmpFolderPath(FilePath workspace) throws IOException, InterruptedException {
+    default String getNodeSpecificTmpFolderPath(FilePath workspace, Launcher launcher) throws IOException, InterruptedException {
         Computer cmp = workspace.toComputer();
         if (cmp == null) {
             throw new IOException(Message.getValue("build.workspace.computer.not.found"));
         }
         String tmpDir = (String) cmp.getSystemProperties().get("java.io.tmpdir");
+
+        // Handle Java temp folder path discrepancy for Windows.
+        if (!launcher.isUnix() && tmpDir.charAt(tmpDir.length() - 1) == '\\'){
```
nbhoski, line 96: We are already verifying the launcher platform  under getProcessToRunMatlabCommand() method .  Can we bring this to that code instead passing Launcher as an extra parameter? Or else even better way is to handdle it at this line matlabLauncher.cmds(tmpDir + "\\" + runnerScriptName, "\"" + matlabCommand + "\"") if tmpdir includes "\" by default.

nbhoski, line 96: I think We could avoid this Sub string thing by by either by converting it to FilePath and just do getRemote() on it which will give us platform specif path.

src/main/java/com/mathworks/ci/MatlabBuild.java
nbhoski: I think that the difference here java.io.tmpdir return `/tmp`  on Linux and `C:\Users\nbhoski\AppData\Local\Temp\`  on Windows, It adds  the trailing backslash on windows by default but not on Linux.

mcafaro: But as far as I know there is no reason java.io.tmpdir could not return `/tmp/` on Linux. java.io.tmpdir is user configurable.

src/main/java/com/mathworks/ci/MatlabBuild.java
abhargav-mw: Thanks! This worked perfectly. Updated the code.

src/main/java/com/mathworks/ci/MatlabBuild.java
abhargav-mw: I updated the code to avoid trimming string and use FilePath.getRemote() instead. FilePath object invokes 'normalize' function which handles extra file separators cross platform.

src/main/java/com/mathworks/ci/MatlabBuild.java
```diff
@@ -76,16 +77,26 @@ default FilePath getFilePathForUniqueFolder(Launcher launcher, String uniqueName
...
         if (cmp == null) {
             throw new IOException(Message.getValue("build.workspace.computer.not.found"));
         }
-        String tmpDir = (String) cmp.getSystemProperties().get("java.io.tmpdir");
-        return tmpDir;
+        
+        String tmpDirPath = (String) cmp.getSystemProperties().get("java.io.tmpdir");
+
+        // Invoke FilePath.normalize for clean file path.
+        FilePath tmpDir = new FilePath(new File(tmpDirPath));
```
nbhoski, line 98: I would recommend to test the behaviour using remote agent once since we are using File() object which represents the local file .

abhargav-mw: I tweaked implementation a bit to get channel from computer and create a FilePath. I confirmed this works both on master and remote agents.

On Master :
jenkinOnMaster

On Remote Agent :
jenkinOnRemoteAgent

src/main/java/com/mathworks/ci/MatlabBuild.java
```diff
@@ -76,16 +77,26 @@ default FilePath getFilePathForUniqueFolder(Launcher launcher, String uniqueName
...
         
         String tmpDir = getNodeSpecificTmpFolderPath(workspace);
-        return new FilePath(launcher.getChannel(), tmpDir+"/"+uniqueName);
+        String fileSeperator = "/";
+
+        if (!launcher.isUnix()){
+            fileSeperator = "\\";
+        }
+
+        return new FilePath(launcher.getChannel(), tmpDir + fileSeperator + uniqueName);
```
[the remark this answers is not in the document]
abhargav-mw: Reverted changes to original implementation as File Path cleans up path anyway.

nbhoski:
Also to note with File() object use none of our unit test failed. We should figure out if we could add unit test which would catch this.

Issue #121, opened 2020-07-22
runMATLABCommand step has timeout of 5 minutes


geerligs:
Hello,

When using the runMATLABCommand step I experience two issues:
- the runMATLABCommand step seems to have a timeout of 5 minutes.
- when having a timeout on the runMATLABCommand step the job is not killed automatically. I need to perform a "hard kill"on Jenkins

Can you confirm these observations? I am running version 2.1.0 of the plugin.

### Timeout of 5 minutes on runMATLABCommand step

To test this I have run the pipeline below. This will set the timeout to 10 minutes and runs a Matlab script which will pause for 6 minutes. 
```
// Declarative Pipeline
pipeline {
    agent any
    options { timestamps() }
    stages {
        stage('test timing 2015'){
            environment {
        			PATH = "C:\\Program Files\\MATLAB\\R2015b\\bin;${PATH}"   // Windows agent
        	}
        	options {
                timeout(time: 10, unit: "MINUTES")
        	}
            steps {
                runMATLABCommand "pause(6*60); ver;pwd"
            } 
        }
    }
}
```
I expect the job to run for 6 minutes and print `ver` and `pwd`. However the job times out after 5 minutes. The Jenkins console shows:
```
11:41:37  Timeout set to expire in 10 min
[Pipeline] {
[Pipeline] withEnv
[Pipeline] {
[Pipeline] runMATLABCommand
11:41:37  Generating MATLAB script with content:
11:41:37  pause(6*60); ver;pwd
11:41:37  
11:41:37  #################### Starting command output ####################
11:41:37  [d5354205-f476-4c03-b2ff-ac097c28ad0a] $ cmd.exe /C C:\Users\han\AppData\Local\Temp\\d5354205-f476-4c03-b2ff-ac097c28ad0a\run_matlab_command.bat "command_f0d95434_4742_42af_a7a3_76ed0ab5286f"
11:41:39   
11:41:39  To get started, type one of these: helpwin, helpdesk, or demo.
11:41:39  For product information, visit www.mathworks.com.
11:41:39   
11:41:40  Warning: Unable to locate a personal folder for $documents\MATLAB
11:41:40  Warning: Unable to locate a personal folder for $documents\MATLAB
11:46:37  ^CTerminate batch job (Y/N)? 
11:46:38  null
```
When replacing the runMATLABCommand step aby a bat-command the job will run for the requested 6 minutes:
[...]

PR #123, opened 2020-07-30
Fixed timeout issue for Pipeline steps 

nbhoski:
* Fixing Timeout issue for Both Pipeline steps.
* Using `SynchronousNonBlockingStepExecution`  class which does not interrupt the process after 5 minutes.

### DIFF
--- a/src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
+++ b/src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
@@ -2,7 +2,7 @@ package com.mathworks.ci;
 
 import java.io.IOException;
 import org.jenkinsci.plugins.workflow.steps.StepContext;
-import org.jenkinsci.plugins.workflow.steps.StepExecution;
+import org.jenkinsci.plugins.workflow.steps.SynchronousNonBlockingStepExecution;
 import hudson.EnvVars;
 import hudson.FilePath;
 import hudson.Launcher;
@@ -10,7 +10,7 @@ import hudson.Launcher.ProcStarter;
 import hudson.model.Result;
 import hudson.model.TaskListener;
 
-public class MatlabCommandStepExecution extends StepExecution implements MatlabBuild {
+public class MatlabCommandStepExecution extends SynchronousNonBlockingStepExecution<Void> implements MatlabBuild {
     
     private static final long serialVersionUID = 1957239693658914450L;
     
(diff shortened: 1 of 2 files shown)


src/main/java/com/mathworks/ci/MatlabRunTestsStepExecution.java
```diff
@@ -46,10 +46,8 @@ public boolean start() throws Exception {
 
         getContext().setResult((res == 0) ? Result.SUCCESS : Result.FAILURE);
         
-        getContext().onSuccess(true);
-        
         //return false represents the asynchronous run. 
-        return false;
+        return null;
```
acampbel, line 50: My Java-Fu is quite old, but why is the `Void` have a capital V here? And also why `return null` over just `return`?

nbhoski: Its just alias for actual void. SynchronousNonBlockingStepExecution uses parameters for return types i choose not to return anything here hence its [Void](https://docs.oracle.com/javase/7/docs/api/java/lang/Void.html#:~:text=Class%20Void&text=The%20Void%20class%20is%20an,representing%20the%20Java%20keyword%20void.) since Void is uninstantiable object i just return null.

acampbel: Sounds good. Now I see the generic parameter in the extends line. Tests? What failing test would have alerted us to this?

nbhoski: We will have to add integration-test for this which would run against actual MATLAB instance. Unit tests may be a challenge though as nothing on UI or log level that could be verified. I can rather keep a test along side unit tests and make it disabled for unit run.

acampbel: @SangaviJayagopi Lets include this as a requirement for MATLAB in the loop integration testing

src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
```diff
@@ -47,10 +47,8 @@ public boolean start() throws Exception {
 
         getContext().setResult((res == 0) ? Result.SUCCESS : Result.FAILURE);
         
-        getContext().onSuccess(true);
-        
         //return false represents the asynchronous run.
```
mcafaro, line 50: Comment no longer valid

src/main/java/com/mathworks/ci/MatlabRunTestsStepExecution.java
```diff
@@ -46,10 +46,8 @@ public boolean start() throws Exception {
 
         getContext().setResult((res == 0) ? Result.SUCCESS : Result.FAILURE);
         
-        getContext().onSuccess(true);
-        
         //return false represents the asynchronous run.
```
mcafaro, line 49: Comment no longer valid

nbhoski: Done.

nbhoski:
Closes #121

PR #130, opened 2020-08-07
Genscript changes

nbhoski:
* Leverages genscript package to generate MATLAB runner script file dynamically.
* Existing runMatlabTests.m file will now only contain logic to generate the script and run based on artifact parameters


src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -238,13 +238,19 @@ private synchronized int execMatlabCommand(FilePath workspace, Launcher launcher
...
-                    constructCommandForTest(getInputArguments()), uniqueTmpFldrName);
+                    constructCommandForTest(getInputArguments(), genScriptLocation.getRemote()),
+                    uniqueTmpFldrName);
             
-            // Copy MATLAB scratch file into the workspace.
-            FilePath targetWorkspace = new FilePath(launcher.getChannel(), workspace.getRemote());
+            // Copy MATLAB scratch file into the workspace.      
             copyFileInWorkspace(MatlabBuilderConstants.MATLAB_TESTS_RUNNER_RESOURCE,
-                    MatlabBuilderConstants.MATLAB_TESTS_RUNNER_TARGET_FILE, targetWorkspace);
+                    MatlabBuilderConstants.MATLAB_TESTS_RUNNER_TARGET_FILE, workspace);
```
mcafaro, line 250: Can we figure out a way not to copy anything to the workspace? Part of the recent work with scriptgen was to eliminate the need to modify the user's workspace. It seems counterproductive to have done all that work and then still put a runner script in the workspace.

The other plugins handle this by running all the code you have in your runner script as the command passed to run_matlab_command. You could also imagine using the function handle trick as well.

nbhoski: Done. using UUID approach with `test_runner` as prefix for the file and keeping file in temp folder.

src/main/java/com/mathworks/ci/MatlabRunTestsStepExecution.java
```diff
@@ -58,10 +58,18 @@ private synchronized int execMatlabCommand(FilePath workspace, Launcher launcher
             TaskListener listener, EnvVars envVars) throws IOException, InterruptedException {
         final String uniqueTmpFldrName = getUniqueNameForRunnerFile();  
         try {
-            ProcStarter matlabLauncher = getProcessToRunMatlabCommand(workspace, launcher, listener, envVars,
-                    envVars.expand(getCommand()), uniqueTmpFldrName);
+            FilePath genScriptLocation =
+                    getFilePathForUniqueFolder(launcher, uniqueTmpFldrName, workspace);
+            final String cmdPrefix = "addpath(genpath('" + genScriptLocation.getRemote() + "')); ";
```
mcafaro, line 63: `genpath` should not be needed. Just the top level path of scriptgen needs to be on the path.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -259,11 +258,12 @@ private synchronized int execMatlabCommand(FilePath workspace, Launcher launcher
...
-
-    public String constructCommandForTest(String inputArguments) {
-        final String matlabFunctionName =
-                FilenameUtils.removeExtension(MatlabBuilderConstants.MATLAB_TESTS_RUNNER_TARGET_FILE);
-        final String runCommand = "exit(" + matlabFunctionName + "(" + inputArguments + "))";
+    
+    public String constructCommandForTest(String inputArguments, FilePath scriptPath) {
+        final String matlabFunctionName = MatlabBuilderConstants.MATLAB_TEST_RUNNER_FILE_PREFIX
+                + scriptPath.getBaseName().replaceAll("-", "_");
+        final String runCommand = "addpath(genpath('" + scriptPath.getRemote() + "')); "
```
mcafaro, line 265: Is `genpath` necessary?

src/main/java/com/mathworks/ci/MatlabRunTestsStepExecution.java
mcafaro, line 63: Don't forget any time you're injecting a string value into MATLAB code, you have to escape single quotes.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
mcafaro, line 265: Don't forget any time you're injecting a string value into MATLAB code, you have to escape single quotes.

pom.xml
```diff
@@ -126,6 +126,18 @@
...
 						</configuration>
 					</execution>
+					<execution>
+						<id>get-matlab-gen-script</id>
+						<phase>validate</phase>
+						<goals>
+							<goal>wget</goal>
+						</goals>
+						<configuration>
+							<url>https://ssd.mathworks.com/supportfiles/ci/matlab-script-generator/v0/matlab-script-generator.zip</url>
```
mcafaro, line 136: How can we keep the caching issue from happening again?

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/runMatlabTests.m
```diff
@@ -1,6 +1,6 @@
-%Copyright 2019-2020 The MathWorks, Inc.
+%Copyright 2020 The MathWorks, Inc.
 
-function failed = runMatlabTests(varargin)
+function runMatlabTests(varargin)
```
mcafaro, line 3: Is this function necessary? The reason I ask, is it's going to add to the users stack trace. 

Could you just run the contents as the command itself? That's not totally ideal either though because then we can get into parsing trouble. I think maybe we should discuss.

pom.xml
nbhoski: Done.

src/main/java/com/mathworks/ci/MatlabRunTestsStepExecution.java
nbhoski: Done.

nbhoski: Thats good point :P . done

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
nbhoski: Done.

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/runMatlabTests.m
nbhoski: To add to it this file will anyways get deleted in either case. May be stack trace, in that case, would be helpful. will discuss this i will keep it open till we conclude.

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/runMatlabTests.m
```diff
@@ -14,175 +14,25 @@
...
-
-function issuePDFReportUnsupportedWarning
-warning('MATLAB:testArtifact:pdfReportNotSupported', ...
-    'Producing a test report in PDF format is not supported in the current MATLAB release.');
-
-function issueExportSTMResultsUnsupportedWarning
-warning('MATLAB:testArtifact:cannotExportSimulinkTestManagerResults', ...
-    ['Unable to export Simulink Test Manager results. This feature ', ...
-    'requires a Simulink Test license and is supported only in MATLAB R2019a or a newer release.']);
+srcFldr = getenv('WORKSPACE');
```
mcafaro, line 24: You should not need to specify `SourceFolder` to achieve the current Jenkins Plugin behavior.

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/runMatlabTests.m
mcafaro: Nikhil and I discussed this offline. 

The pro of writing a file to disk is that we do not have to worry about parsing issues with injecting user provided text into a shell command input. The con is that the file becomes part of the users stack trace.

Neither felt like a clear win, so we will keep this as is for now.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -259,11 +258,12 @@ private synchronized int execMatlabCommand(FilePath workspace, Launcher launcher
...
-
-    public String constructCommandForTest(String inputArguments) {
-        final String matlabFunctionName =
-                FilenameUtils.removeExtension(MatlabBuilderConstants.MATLAB_TESTS_RUNNER_TARGET_FILE);
-        final String runCommand = "exit(" + matlabFunctionName + "(" + inputArguments + "))";
+    
+    public String constructCommandForTest(String inputArguments, FilePath scriptPath) {
+        final String matlabFunctionName = MatlabBuilderConstants.MATLAB_TEST_RUNNER_FILE_PREFIX
+                + scriptPath.getBaseName().replaceAll("-", "_");
+        final String runCommand = "addpath('" + scriptPath.getRemote().replaceAll("'", "''")
```
mcafaro, line 265: By using `addpath`, the temp path is going to be on the user's path throughout the entire run. We try to avoid modifying the path as much as possible. Consider using `cd` to go to the folder containing the script and then `cd` back in the script. I think that's how you are handling this in Run MATLAB Command.

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/runMatlabTests.m
nbhoski: Yes and we shall continue persuading this till we conclude on a robust solution  I have raise #133 to track the same

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/runMatlabTests.m
nbhoski: Done.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
nbhoski: I have to ad path anyways in the script file to get testScript.writeToText() and run(testScript) recognized. Changed it accordingly.

PR #143, opened 2020-08-27
Support to write test runner script instead copying in worksapce.

### LINKED ISSUE #134: Use a script for runMatlabTest instead of a function
With the integration of `genscript` into the plugin (#130), the `runMatlabTest` function does not do much other than pass through name-value pairs to `genscript`. By keeping `runMatlabTest` a function, we are required to pass user-provided values to it through the system shell (e.g. `run_matlab_command runMatlabTest(USER_PROVIDED_VALUES)`). These user provided values will be parsed by the system shell before they make it into the function. This parsing can and will introduce unexpected issues.

Instead of passing these user-provided values to a function, the plugin should write them directly into a templated script. Then we can call `run_matlab_command` on the script and not worry about the system shell parsing the user provided values.

The templated script would look something like this:

```matlab
testScript = genscript('Test',...
   'PDFTestReport', ${PDF_TEST_REPORT},...
   'TAPTestResults', ${TAP_TEST_RESULTS},...
   'JUnitTestResults', ${JUNIT_TEST_RESULTS},...
   'SimulinkTestResults', ${SIMULINK_TEST_RESULTS},...
   'CoberturaCodeCoverage', ${COBERTURA_CODE_COVERAGE},...
   'CoberturaModelCoverage', ${COBERTURA_MODEL_COVERAGE});

disp('Running MATLAB script with contents:\n');
disp(testScript.Contents);
fprintf('___________________________________\n\n');
run(testScript);
```

Then the plugin will need to find/replace the template placeholders with appropriate values when writing it to disk.

nbhoski:
This PR closes #134 
* Writing the entire test runner script runtime instead copying it from the resource bundle
* genscript() method receives only those artifacts as parameters which were selected on UI.
* Console output does not show a long list of arguments as the test runner is just a script file now and not a method.


src/test/java/com/mathworks/ci/RunMatlabTestsBuilderTest.java
```diff
@@ -221,10 +221,11 @@ public void verifyVerlessThan() throws Exception {
     }
 
     /*
-     * Test to verify appropriate test atrtifact values are passed.
+     * Test to verify appropriate test atrtifact values are passed. Need to 
+     * include in integration test.
      */
 
-    @Test
+
```
abhargav-mw, line 228: This test didn't run in the current qualification. Need to add the @Test annotation.

src/test/java/com/mathworks/ci/RunMatlabTestsStepTest.java
```diff
@@ -67,15 +67,15 @@ public void verifyOnslave() throws Exception {
     }
 
     /*
-     * Verify artifact path is correct.
+     * Verify artifact path is correct. Need to move this to integration test.
      */
 
-    @Test
+    
     public void verifyArtifactPath() throws Exception {
```
abhargav-mw, line 74: This test didn't run in the current qualification. Need to add the @ test annotation.

abhargav-mw (changes requested):

src/test/java/com/mathworks/ci/RunMatlabTestsBuilderTest.java
```diff
@@ -221,10 +221,11 @@ public void verifyVerlessThan() throws Exception {
     }
 
     /*
-     * Test to verify appropriate test atrtifact values are passed.
+     * Test to verify appropriate test atrtifact values are passed. Need to 
+     * include in integration test.
```
nbhoski, line 225: This is intentional. The test behavior changes after new changes and is updated as per new behavior. however, i did not want to remove it from test set. this should be later included in the integration test set once integ tests are part of repo hence updated the comment accordingly and skipped it from unit run

src/test/java/com/mathworks/ci/RunMatlabTestsBuilderTest.java
nbhoski: This is intentional. The test behavior changes after new changes and is updated as per new behavior. however, i did not want to remove it from test set. this should be later included in the integration test set once integ tests are part of repo hence updated the comment accordingly and skipped it from unit run

src/test/java/com/mathworks/ci/RunMatlabTestsStepTest.java
nbhoski: Same as earlier.

src/main/java/com/mathworks/ci/MatlabBuilderConstants.java
```diff
@@ -34,4 +33,10 @@
     
     //Test runner file prefix 
     static final String MATLAB_TEST_RUNNER_FILE_PREFIX = "test_runner_";
+    
+    // MATLAB runner script
+    static final String TEST_RUNNER_SCRIPT = "testScript = genscript(${PARAMS});\n" + "\n"
+            + "disp('Running MATLAB script with content:\\n');\n"
```
mcafaro, line 39: I don't think `disp` recognizes special characters like `\n`. `disp` will automatically add a new line though, so you probably don't need `\n` here.

src/main/java/com/mathworks/ci/RunMatlabTestsStep.java
```diff
@@ -121,7 +121,9 @@ public String getDisplayName() {
     
     private String getInputArgs() {
         final List<String> inputArgs = new ArrayList<>();
-        final Map<String, String> args = getMatlabArgs();
+        final Map<String, String> args = getGenscriptArgs();
+        
+        inputArgs.add("'Test'");
```
mcafaro, line 126: You have a check for `inputArgs.isEmpty()` below here that seems like it would never occur given this line.

src/main/java/com/mathworks/ci/MatlabBuild.java
```diff
@@ -98,12 +98,12 @@ default String getUniqueNameForRunnerFile() {
...
-        // Copy MATLAB scratch file in temp.
-        copyFileInWorkspace(MatlabBuilderConstants.MATLAB_TESTS_RUNNER_RESOURCE,
-                MatlabBuilderConstants.MATLAB_TEST_RUNNER_FILE_PREFIX
-                        + tmpFldr.getBaseName().replaceAll("-", "_") + ".m",
-                tmpFldr);
+    default void prepareTmpFldr(FilePath tmpFldr, String runnerScript) throws IOException, InterruptedException {
+        // Write MATLAB scratch file in temp folder.
+        FilePath scriptFile =
+                new FilePath(tmpFldr, MatlabBuilderConstants.MATLAB_TEST_RUNNER_FILE_PREFIX
+                        + tmpFldr.getBaseName().replaceAll("-", "_") + ".m");
```
mcafaro, line 105: `.getBaseName().replaceAll("-","_")` seems like a common occurrence in the code. You may consider performing this once and storing the result in a variable or putting this in a function or something.

MATLAB has more restriction on script names than just not allowing "-" (you can't start it with a number for instance, see [isvarname](https://www.mathworks.com/help/matlab/ref/isvarname.html)). You happen to know that `tmpFldr` isn't going to get a folder name that starts with a number, but it's best to avoid this kind of shared knowledge throughout the code if possible. You may eventually forget `prepareTmpFldr` has this precondition and new contributors won't know it to begin with.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -278,6 +278,8 @@ private String getInputArguments() {
                 new ArrayList<Artifact>(Arrays.asList(getPdfReportArtifact(), getTapArtifact(),
                         getJunitArtifact(), getStmResultsArtifact(), getCoberturaArtifact(),
                         getModelCoverageArtifact()));
+        
+        inputArgsList.add("'Test'");
 
         for (Artifact artifact : artifactList) {
             artifact.addFilePathArgTo(args);
```
mcafaro, line 285: Should we be expanding environment variables in the field values before they we write them to the script file?

I think we were expanding env variables before, so it would be a change of behavior if we didn't.

src/main/java/com/mathworks/ci/MatlabBuilderConstants.java
nbhoski: Done.

src/main/java/com/mathworks/ci/RunMatlabTestsStep.java
nbhoski: Doe.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
nbhoski: I don't think we were expanding test artifact values before. We were expanding it for Command text box and MATLAB root input text box.  Matrix build maintains individual workspace for each axis that we add hence artifacts will always be unique in each workspace. expansion in artifact text would just allow substituting Env names for freestyle or assigning the same name as of axis in Matrix (Which i feel is not strong use case). However, we could discuss this if we should support expansion for test artifacts.

src/main/java/com/mathworks/ci/MatlabBuild.java
nbhoski: This was good point @mcafaro . Thanks for this perspective. I have extracted this method out in isolation which knows the details of MATLAB specific file names. and others need not be aware of this.

mw-akumar (approved):

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
mcafaro: On line 69 of `MatlabRunTestsStepExecution` we used to have:

`matlabFunctionName+ "("+envVars.expand(getCommand()+")")`

And `getCommand()` used to return something like `addpath(...); runMatlabTests('PDFTestReport', 'SOME_FILE_PATH', ...)` right?

Which means if the user specified `SOME_FILE_PATH` as a Jenkins variable (e.g. `${MY_FILE_PATH}`), I think we would have expanded it.

Not sure this is a significant use-case but it seems like it would be a change of behavior if we don't expand them now.

nbhoski: Done.

PR #144, opened 2020-09-02
GitHub issue 138

### LINKED ISSUE #138: In a Multi-Configuration project, if a MATLAB version does not exist in the system, no error or warning generated
For example, if I dont have version R2018a in the system, in a Free-style project, I get a warning message in the UI Configuration page "Unable to find MATLAB on this machine using the specified location. The specified root folder might contain MATLAB on a different build agent"

In a Multi-config project, no such error or warning at the end of the build.

nbhoski:
This PR closes #138 
* UseMatlabVersion now verifies if MATLAB executable is present at a given location.
* If MATLAB executable could not be found then throws Error on Jenkins console and fails the build.
* Error is thrown even if Already some MATLAB is in the Path and user provides MATLAB version explicitly using UseMatlabVersion
* If MATLAB is not explicitly provided using UseMatlabVersion then default MATLAB on the Path will be picked.
* Added dummy executable files in test resources.
* Updated the failed test and added new to verify the functionality.


src/main/resources/config.properties
```diff
@@ -12,6 +12,7 @@ Builder.matlab.test.support.error = To run tests with the Jenkins plugin, use MA
 builder.matlab.automatictestoption.display.name = Automatic
 builder.matlab.customcommandoption.display.name = Custom
 Releaseinfo.matlab.version.not.found.error = Error finding MATLAB release for given MATLAB root. Verify MATLAB root path.
+matlab.not.found.error = Error finding MATLAB at given location. Verify MATLAB root path.
```
mw-hrastega, line 15: Can we use this text?

Unable to launch MATLAB from the specified location. Verify the path to MATLAB root folder.

nbhoski: Done.

mw-hrastega (approved):

src/main/java/com/mathworks/ci/UseMatlabVersionBuildWrapper.java
```diff
@@ -133,18 +134,20 @@ public synchronized void setUp(Context context, Run<?, ?> build, FilePath worksp
...
+                getLocalMatlab() + "/bin/" + getNodeSpecificExecutable(launcher));
+
+        if (!matlabExecutablePath.exists()) {
+            throw new MatlabNotFoundError(Message.getValue("matlab.not.found.error"));
+        }
         // Add "matlabroot" without bin as env variable which will be available across the build.
         context.env("matlabroot", getLocalMatlab());
         // Add matlab bin to path to invoke MATLAB directly on command line.
-        context.env("PATH+matlabroot", getLocalMatlab() + nodeSpecificFileSep + "bin");     
+        context.env("PATH+matlabroot", getLocalMatlab() + "/bin");
```
mw-akumar, line 147: Would it be better to just matlabExecutablePath.parent() ? This way we are using the FilePath abstraction and reusing our previously computed matlabExecutablePath?

src/main/java/com/mathworks/ci/UseMatlabVersionBuildWrapper.java
```diff
@@ -133,18 +134,20 @@ public synchronized void setUp(Context context, Run<?, ?> build, FilePath worksp
         // Set Environment variable
 
         setEnv(initialEnvironment);
-        String nodeSpecificFileSep = getNodeSpecificFileSeperator(launcher);
```
mcafaro, line 136: Why don't we need to be concerned with the node specific file separator now?

src/main/resources/config.properties
```diff
@@ -12,6 +12,7 @@ Builder.matlab.test.support.error = To run tests with the Jenkins plugin, use MA
 builder.matlab.automatictestoption.display.name = Automatic
 builder.matlab.customcommandoption.display.name = Custom
 Releaseinfo.matlab.version.not.found.error = Error finding MATLAB release for given MATLAB root. Verify MATLAB root path.
+matlab.not.found.error = Unable to launch MATLAB from the specified location. Verify the path to MATLAB root folder.
```
mcafaro, line 15: @mw-hrastega should take a look at this.

mcafaro (approved):
Overall looks good. Just a question about why we can always use a "/" file sep now and couldn't before. Houman should probably take a look at the error message as well.

src/main/java/com/mathworks/ci/UseMatlabVersionBuildWrapper.java
nbhoski: I learnt it late that FilePath handles "/" nd after initial draft did not get chance to touch this file :)

src/main/java/com/mathworks/ci/UseMatlabVersionBuildWrapper.java
nbhoski: Done.

PR #148, opened 2020-09-08
Add Source Folder option

### LINKED ISSUE #63: Automatic mode for non-project based repositories
I don't see any calls to `addpath` in `runMatlabTests.m`. How do your source files get added to the search path if you have a non-project based repository? Is this not currently supported?

abhargav-mw:
This PR contains changes for :
- Adding Source Folder option to Pipeline Project.
- Adding Source Folder option to FreeStyle Project.
- Adding unit test for Source Folder option.
- Adding unit test to validate RunMatlabTestBuilder Artifacts persistence after Restarting Jenkins.

This PR closes #63

### DIFF
--- a/src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
+++ b/src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
@@ -11,18 +11,12 @@ package com.mathworks.ci;
 import java.io.IOException;
 import java.util.*;
 import javax.annotation.Nonnull;
+import hudson.*;
+import hudson.model.*;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.DataBoundSetter;
 import org.kohsuke.stapler.StaplerRequest;
-import hudson.EnvVars;
-import hudson.Extension;
-import hudson.FilePath;
-import hudson.Launcher;
 import hudson.Launcher.ProcStarter;
-import hudson.model.AbstractProject;
-import hudson.model.Result;
-import hudson.model.Run;
-import hudson.model.TaskListener;
 import hudson.tasks.BuildStepDescriptor;
 import hudson.tasks.Builder;
 import jenkins.tasks.SimpleBuildStep;
@@ -47,15 +41,14 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
     private Artifact stmResultsArtifact = new NullArtifact();
     private Artifact modelCoverageArtifact = new NullArtifact();
     private Artifact pdfReportArtifact = new NullArtifact();
-   
+    private SourceFolder sourceFolder;
+
     @DataBoundConstructor
     public RunMatlabTestsBuilder() {
 
     }
 
-
     // Getter and Setters to access local members
-
 
     @DataBoundSetter
     public void setTapArtifact(TapArtifact tapArtifact) {
@@ -86,7 +79,13 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
     public void setPdfReportArtifact(PdfArtifact pdfReportArtifact) {
         this.pdfReportArtifact = pdfReportArtifact;
     }
-    
+
+    @DataBoundSetter
+    public void setSourceFolder(SourceFolder sourceFolder) {
+        this.sourceFolder = sourceFolder;
+    }
+
+
     public String getTapReportFilePath() {
         return this.getTapArtifact().getFilePath();
     }      
@@ -133,6 +132,10 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
     
     public String getPdfReportFilePath() {
         return this.getPdfReportArtifact().getFilePath();
+    }
+
+    public SourceFolder getSourceFolder() {
+        return this.sourceFolder;
     }
 
     private Artifact getArtifactObject(boolean isChecked, Artifact returnVal)  {
@@ -284,6 +287,12 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
             artifact.addFilePathArgTo(args);
         }
 
+        // Add source folder options to argument
+        SourceFolder sf = getSourceFolder();
+        if(sf != null && sf.getSourceFolderPaths().size() != 0){
+            sf.addFilePathArgTo(args);
+        }
+
         args.forEach((key, val) -> inputArgsList.add("'" + key + "'" + "," + "'" + val + "'"));
 
         return String.join(",", inputArgsList);
@@ -300,6 +309,7 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
      * 7Csort:date/jenkinsci-dev/AFYHSG3NUEI/UsVJIKoE4B8J
      * 
      */
+
     public static class PdfArtifact extends AbstractArtifactImpl {
 
         private static final String PDF_TEST_REPORT = "PDFTestReport";
--- a/src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
+++ b/src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
@@ -15,6 +15,8 @@ import java.util.List;
 import java.util.ArrayList;
 import java.util.Optional;
 import java.util.Arrays;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
 import hudson.EnvVars;
 import hudson.Extension;
@@ -37,7 +39,6 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
 
     private int buildResult;
     private EnvVars env;
-    private static final String SOURCE_FOLDER = "SourceFolder";
 
     // Make all old values transient which protects them writing back on disk.
     private transient boolean tapChkBx;
@@ -282,6 +283,15 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
         return runCommand;
     }
 
+    private String getCellArrayFrmList(List<String> listOfStr){
+        // Ignore empty string values in the list
+        Predicate<String> isEmpty = String::isEmpty;
+        Predicate<String> isNotEmpty = isEmpty.negate();
+        List<String> filteredListOfStr = listOfStr.stream().filter(isNotEmpty).collect(Collectors.toList());
+        filteredListOfStr.replaceAll(val -> "'" + val.replaceAll("'", "''") + "'");
+        return "{" + String.join(",", filteredListOfStr) + "}";
+    }
+
     // Concatenate the input arguments
     private String getInputArguments() {
 
@@ -299,19 +309,15 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
             artifact.addFilePathArgTo(args);
         }
 
+        args.forEach((key, val) -> inputArgsList.add("'" + key + "'" + "," + "'" + val + "'"));
+
         // Add source folder options to argument
         SourceFolder sf = getSourceFolder();
         if(sf != null && !sf.getSourceFolderPaths().isEmpty()){
-            sf.addFilePathArgTo(SOURCE_FOLDER, args);
+            sf.addSourceToInputArgs(inputArgsList, getCellArrayFrmList(sf.getSourceFolderPaths().stream()
+                    .map(SourceFolderPaths::getSrcFolderPath)
+                    .collect(Collectors.toList())));
         }
-
-        args.forEach((key, val) -> {
-            if(key.equals(SOURCE_FOLDER)){
-                inputArgsList.add("'" + key + "'" + "," + val);
-            }else{
-                inputArgsList.add("'" + key + "'" + "," + "'" + val + "'");
-            }
-        });
 
         return String.join(",", inputArgsList);
     }
--- a/src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
+++ b/src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
@@ -15,7 +15,6 @@ import java.util.List;
 import java.util.ArrayList;
 import java.util.Optional;
 import java.util.Arrays;
-import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
 import hudson.EnvVars;
@@ -283,15 +282,6 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
         return runCommand;
     }
 
-    private String getCellArrayFrmList(List<String> listOfStr){
-        // Ignore empty string values in the list
-        Predicate<String> isEmpty = String::isEmpty;
-        Predicate<String> isNotEmpty = isEmpty.negate();
-        List<String> filteredListOfStr = listOfStr.stream().filter(isNotEmpty).collect(Collectors.toList());
-        filteredListOfStr.replaceAll(val -> "'" + val.replaceAll("'", "''") + "'");
-        return "{" + String.join(",", filteredListOfStr) + "}";
-    }
-
     // Concatenate the input arguments
     private String getInputArguments() {
 
@@ -311,10 +301,13 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
 
         args.forEach((key, val) -> inputArgsList.add("'" + key + "'" + "," + "'" + val + "'"));
 
-        // Add source folder options to argument
+        /*
+        * Add source folder options to argument.
+        * For source folder we create a MATLAB cell array and add it to input argument list.
+        * */
         SourceFolder sf = getSourceFolder();
         if(sf != null && !sf.getSourceFolderPaths().isEmpty()){
-            sf.addSourceToInputArgs(inputArgsList, getCellArrayFrmList(sf.getSourceFolderPaths().stream()
+            sf.addSourceToInputArgs(inputArgsList, Utilities.getCellArrayFrmList(sf.getSourceFolderPaths().stream()
                     .map(SourceFolderPaths::getSrcFolderPath)
                     .collect(Collectors.toList())));
         }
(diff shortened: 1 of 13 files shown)


src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -11,18 +11,12 @@
 import java.io.IOException;
 import java.util.*;
 import javax.annotation.Nonnull;
+import hudson.*;
```
mw-akumar, line 14: I am not sure what the best practice in Java is, so we should check that before importing all the symbols.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -284,6 +287,12 @@ private String getInputArguments() {
             artifact.addFilePathArgTo(args);
         }
 
+        // Add source folder options to argument
```
mw-akumar, line 290: I wonder if this should follow the existing Artifact to arguments pattern instead of standing alone like this. It could be that this particular artifact may need custom behavior when gets converted to arguments.

src/main/java/com/mathworks/ci/RunMatlabTestsStep.java
```diff
@@ -142,6 +153,11 @@ private String getInputArgs() {
         args.put("SimulinkTestResults", getTestResultsSimulinkTest());
         args.put("CoberturaCodeCoverage", getCodeCoverageCobertura());
         args.put("CoberturaModelCoverage", getModelCoverageCobertura());
+        if(getSourceFolder() != null && getSourceFolder().size() != 0){
```
mw-akumar, line 156: since fixNull is used, do we need the null check?

mw-akumar, line 156: Using isEmpty is more idiomatic that size != 0

src/main/java/com/mathworks/ci/SourceFolder.java
```diff
@@ -0,0 +1,54 @@
+package com.mathworks.ci;
+
+/**
+ * Copyright 2019-2020 The MathWorks, Inc.
```
mw-akumar, line 4: copyright 2020?

mw-akumar (changes requested):

src/main/java/com/mathworks/ci/SourceFolderPaths.java
```diff
@@ -0,0 +1,37 @@
+package com.mathworks.ci;
+
+/**
+ * Copyright 2019-2020 The MathWorks, Inc.
```
nbhoski, line 4: Copyright 2020

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/help-sourceFolder.html
```diff
@@ -0,0 +1,4 @@
+<div>
+    <p>Optionally specify the location of the folder containing source code, relative to the project root folder. The specified folder and its subfolders are added to the top of the MATLAB search path. To generate a code coverage report, MATLAB uses only the source code in the specified folder and its subfolders. </p>
```
nbhoski, line 2: add @mw-hrastega in review for these help text

nbhoski (changes requested):

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/help-sourceFolder.html
mw-hrastega, line 2: This looks good. I would remove "Optionally". I would also use "folders" instead of folder because we can have multiple folders.
Is source folder useful for Simulink Test model coverage report as well? If so, please drop the word "code". So, the text would become:

Specify the location of folders containing source code, relative to the project root folder. The specified folders and their subfolders are added to the top of the MATLAB search path. To generate a coverage report, MATLAB uses only the source code in the specified folders and their subfolders.

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/help-sourceFolder.html
```diff
@@ -0,0 +1,4 @@
...
+    <p>Specify a source folder path in the input box. Click on 'Add folder' to add more values.</p>
```
mw-hrastega, line 3: Paste the source folder path in the **File path** box. To add more folders, click **Add folder**.

Please confirm that **File path** and **Add folder** are correct.

mw-hrastega (changes requested):
The UI text is good, overall. But some changes are necessary.

src/main/java/com/mathworks/ci/SourceFolder.java
```diff
@@ -0,0 +1,54 @@
...
+
+    public List<SourceFolderPaths> getSourceFolderPaths() {
+        return this.sourceFolderPaths;
+    }
+
+    public  void addFilePathArgTo(Map<String, String> inputArgs) {
+        // Concatenate all source folders to a single ";" separated string
+        inputArgs.put(SOURCE_FOLDER, this.sourceFolderPaths.stream()
+                .map(SourceFolderPaths::getSrcFolderPath)
+                .collect(Collectors.joining(";")));
```
mcafaro, line 44: Just note that using this approach if someone provides a path including a colon, like with an absolute Windows path including a drive letter (e.g. C:\my\folder), the folder will be split into two (e.g. "C" and "\my\folder"). I know the help text says you must use a relative path but I think some users are bound to try an absolute path anyway.

We accepted this limitation previously because the only way a user could provide multiple paths in other systems is with a single string and we wanted to support both ":" and ";" as a path separator. With Jenkins, we don't have that limitation, so perhaps we should construct a cell array instead?

mcafaro (changes requested):
+1 to most of the other comments already made by @mw-akumar, @nbhoski, and @mw-hrastega.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
abhargav-mw: There's a slight difference in notion of an Artifact in this class and Source Folder option, that is the Artifact is generated and Source Folder is a configuration for running test and collecting coverage. Moreover, the implementation for Source Folder differs from Artifact as for it (source folder) we need a describable class. If the source folder variable is initialized with any object, it will show up as Selected in UI (because of optionalProperty jelly tag) hence a Null Object Pattern probably wont be fit for this.

src/main/java/com/mathworks/ci/RunMatlabTestsStep.java
abhargav-mw: Changed to isEmpty

src/main/java/com/mathworks/ci/SourceFolder.java
abhargav-mw: Changed to 2020

src/main/java/com/mathworks/ci/RunMatlabTestsStep.java
abhargav-mw: Fixed

src/main/java/com/mathworks/ci/SourceFolder.java
```diff
@@ -0,0 +1,54 @@
...
+    private List<SourceFolderPaths> sourceFolderPaths = new ArrayList<>();
+    private static final String SOURCE_FOLDER = "SourceFolder";
+
+    @DataBoundConstructor
+    public SourceFolder() {
+
+    }
+
+    @DataBoundSetter
+    public void setSourceFolderPaths(List<SourceFolderPaths> sourceFolderPaths) {
```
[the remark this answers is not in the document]
abhargav-mw: Done

src/main/java/com/mathworks/ci/SourceFolderPaths.java
abhargav-mw: Fixed

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/help-sourceFolder.html
abhargav-mw: Thanks for review. I changed the text as per your suggestion.

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/help-sourceFolder.html
abhargav-mw: Done.

src/main/java/com/mathworks/ci/SourceFolder.java
abhargav-mw: Thank you for pointing this out. Refactored the code to generate a string in cell array format.

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/help-sourceFolder.html
```diff
@@ -0,0 +1,4 @@
+<div>
+    <p>Specify the location of folders containing source code, relative to the project root folder. The specified folders and their subfolders are added to the top of the MATLAB search path. To generate a coverage report, MATLAB uses only the source code in the specified folders and their subfolders.</p>
+    <p>Paste the source folder path in the File path box. To add more folders, click Add folder.</p>
```
mw-hrastega, line 3: Thanks for the changes. We use boldface when mentioning UI text. So, a couple of changes are required in the second paragraph:

File path  -->  **File path**
Add folder --> **Add folder**

mw-hrastega (changes requested):

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/help-sourceFolder.html
abhargav-mw: Done

mw-hrastega (approved):

src/main/java/com/mathworks/ci/RunMatlabTestsStep.java
```diff
@@ -22,16 +22,17 @@
 import hudson.Launcher;
 import hudson.model.Run;
 import hudson.model.TaskListener;
+import hudson.Util;
 
-public class RunMatlabTestsStep extends Step {
+public class RunMatlabTestsStep extends Step implements MatlabBuild{
```
nbhoski, line 27: I would still suggest just for one common method we need not implement MatlabBuild here see if you could scope it to step execution class or within step class itself

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -284,7 +299,19 @@ private String getInputArguments() {
...
 
-        args.forEach((key, val) -> inputArgsList.add("'" + key + "'" + "," + "'" + val + "'"));
+        // Add source folder options to argument
+        SourceFolder sf = getSourceFolder();
+        if(sf != null && !sf.getSourceFolderPaths().isEmpty()){
+            sf.addFilePathArgTo(SOURCE_FOLDER, args);
+        }
+
+        args.forEach((key, val) -> {
+            if(key.equals(SOURCE_FOLDER)){
```
nbhoski, line 309: Would it be good idea if Source folder knew what it adds as parameters to inputArgsList. for example 
 sf.addFilePathArgTo(inputArgsList) . the rationale is the source folder knows what it expects and it provides args in that way.

src/main/java/com/mathworks/ci/RunMatlabTestsStep.java
```diff
@@ -93,6 +94,14 @@ public void setModelCoverageCobertura(String modelCoverageCobertura) {
         this.modelCoverageCobertura = modelCoverageCobertura;
     }
 
+    public List<String> getSourceFolder() {
```
nbhoski, line 97: I did not see the config.jelly changes for this file.

nbhoski (changes requested):

src/main/java/com/mathworks/ci/RunMatlabTestsStep.java
abhargav-mw: Made appropriate changes

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
abhargav-mw: Refactored the code.

src/main/java/com/mathworks/ci/RunMatlabTestsStep.java
abhargav-mw: Refactored to have duplicate methods in RunMatlabTestBuilder and RunMatlabTestStep

src/main/resources/com/mathworks/ci/RunMatlabTestsStep/help-sourceFolder.html
```diff
@@ -0,0 +1,5 @@
+<div>
+    <p>Specify the location of folders containing source code, relative to the project root folder. The specified folders and their subfolders are added to the top of the MATLAB search path. To generate a coverage report, MATLAB uses only the source code in the specified folders and their subfolders.</p>
+    <p>Paste the source folder path in the box in a list format.</p>
+    <p>Example: ["src/folderA", "src/folderB"]</p>
```
mcafaro, line 4: This isn't true anymore, right?

abhargav-mw: Which part? This is true for pipeline. 
This help file is for Pipeline Syntax Snippet Generator section.

mcafaro: Oh I see. Thanks for the clarification.

src/main/java/com/mathworks/ci/RunMatlabTestsStep.java
```diff
@@ -142,6 +156,18 @@ private String getInputArgs() {
...
     }
+
+    private String getCellArrayFrmList(List<String> listOfStr){
+        // Ignore empty string values in the list
+        Predicate<String> isEmpty = String::isEmpty;
+        Predicate<String> isNotEmpty = isEmpty.negate();
+        List<String> filteredListOfStr = listOfStr.stream().filter(isNotEmpty).collect(Collectors.toList());
+        filteredListOfStr.replaceAll(val -> "'" + val.replaceAll("'", "''") + "'");
+        return "{" + String.join(",", filteredListOfStr) + "}";
+    }
```
mcafaro, line 172: Can you avoid repeating this function?

mcafaro (approved):
Looks good to me. Just a note on some repeated code you may be able to refactor.

src/main/resources/com/mathworks/ci/RunMatlabTestsStep/help-sourceFolder.html
mw-hrastega: RE: Paste the source folder path in the box in a list format.

It's not clear what is meant by "in a list format". Can't we remove it?

src/main/java/com/mathworks/ci/RunMatlabTestsStep.java
nbhoski: would be good idea to just start a Util class it would be a good start atleast with one method in it. looks like getting cell array from list may be resused for testfoldrselction later.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
nbhoski: What is it doing differently than earlier ? am i missing something . As i mentioned sf.addFilepathArgTo() adds required arguments in List<> then you may not require the check for key in the forEach

nbhoski (approved):
Few minor comments. rest looks good

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
abhargav-mw: Hi Nikhil,

Could you check my last commit? I have made that change https://github.com/mathworks/jenkins-matlab-plugin/pull/148/commits/86a9bd281249c5aee8ba564aac8090df07ffed6d 

Best,
Anmol

src/main/resources/com/mathworks/ci/RunMatlabTestsStep/help-sourceFolder.html
abhargav-mw: Hi @mw-hrastega , Here(Pipeline Syntax Snippet Generator) the help text is slightly different than the help text in FreeStyle Projects. Here there's a single textbox for entering a list of values for source folder.
I included an example for the same in the help text. 
Could you suggest a better way of putting it?

mw-akumar (approved):
Looks good modulo the comments from Mark about repetition.

src/main/java/com/mathworks/ci/RunMatlabTestsStep.java
abhargav-mw: I had something similar in an earlier commit https://github.com/mathworks/jenkins-matlab-plugin/pull/148/commits/d1c2fca54e6000c51007edcd95b7c18df8cfb6a3 . I'll create a new class 'Utilities' and add this method there.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -289,11 +301,17 @@ private String getInputArguments() {
...
         // Add source folder options to argument
         SourceFolder sf = getSourceFolder();
-        if(sf != null && sf.getSourceFolderPaths().size() != 0){
-            sf.addFilePathArgTo(args);
+        if(sf != null && !sf.getSourceFolderPaths().isEmpty()){
+            sf.addFilePathArgTo(SOURCE_FOLDER, args);
         }
 
-        args.forEach((key, val) -> inputArgsList.add("'" + key + "'" + "," + "'" + val + "'"));
+        args.forEach((key, val) -> {
```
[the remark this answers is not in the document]
abhargav-mw: Sure, I'll add a comment there.

mw-hrastega (approved):

PR #156, opened 2020-09-29
Src folder doc

mw-hrastega:
- Updated all screenshots in the guide and the example.
- Documented the source folder option for web UI-based and Pipeline projects.
- Made tweaks as necessary.


CONFIGDOC.md
```diff
@@ -268,12 +276,13 @@ node {
 | testResultsSimulinkTest 	| Path to export Simulink Test Manager results in MLDATX format (requires Simulink Test license and is supported in MATLAB R2019a or later).<br/>**Example:** `'test-results/results.mldatx'` 	|
 | codeCoverageCobertura   	| Path to write code coverage report in Cobertura XML format.<br/>**Example:** `'code-coverage/coverage.xml'`     	|
 | modelCoverageCobertura  	| Path to write model coverage report in Cobertura XML format (requires Simulink Coverage™ license and is supported in MATLAB R2018b or later).<br/>**Example:** `'model-coverage/coverage.xml'`   	|
+| sourceFolder  	| Locations of the folders containing source code, relative to the project root folder. The specified folders and their subfolders are added to the top of the MATLAB search path. To generate a coverage report, MATLAB uses only the source code in the specified folders and their subfolders. <br/>**Example:** `'source'`<br/>**Example:** `["src/folderA", "src/folderB"]`   	|
```
mcafaro, line 279: Is there a reason to use double quotes over single quotes for the source folder array example? I think either will work but single quotes would be more consistent with the rest of the document:

`['src/folderA', 'src/folderB']`

mcafaro (approved):
Looks good other than a minor suggestion.

Unrelated to this pull, I just noticed the first section says:

"When you define a Jenkins Pipeline, whether in the web UI or with a Jenkinsfile"

I do not believe Jenkins considers the classic web UI a Jenkins Pipeline. A Jenkins Pipeline specifically refers to the use of a Jenkinsfile.

CONFIGDOC.md
```diff
@@ -239,7 +245,8 @@ pipeline {
             steps
             {
                 runMATLABTests(testResultsJUnit: 'test-results/results.xml',
-                               codeCoverageCobertura: 'code-coverage/coverage.xml')
+                               codeCoverageCobertura: 'code-coverage/coverage.xml',
+                               sourceFolder: 'source')
```
abhargav-mw, line 249: Even for a single value being passed to sourceFolder, user needs to include square brackets. Something to the effect of,

sourceFolder: ['source']

else there's a type mismatch error.

abhargav-mw (changes requested):
Looks good!

CONFIGDOC.md
mcafaro: Unfortunate :(. Especially for MATLAB users. Is there a way we can support both scalar and array values?

CONFIGDOC.md
mw-hrastega: Thanks for your comment. Changed it.

CONFIGDOC.md
mw-hrastega: Fixed the examples.

CONFIGDOC.md
mcafaro: Looks good.

nbhoski: Actually it does , Jenkins has separate Pipeline type project like Matrix project other than freestyle
image

CONFIGDOC.md
```diff
@@ -44,21 +45,21 @@ The **Run MATLAB Command** build step enables you to specify MATLAB commands tai
...
 
-The **Run MATLAB Tests** build step enables you to generate different types of test artifacts. To publish the test results, you can use these artifacts with other Jenkins plugins. By default, the plugin assigns a name to each selected artifact and stores it in the `matlabTestArtifacts` folder of the project workspace. You can override the default artifact name and location by specifying a path relative to the project folder in the **File path** box. (If you leave the text box empty, the plugin does not generate an artifact.)  
- 
-![run_matlab_tests](https://user-images.githubusercontent.com/48831250/84545463-54406180-accd-11ea-890e-7a6bb9ee051b.png)
+ With the **Run MATLAB Tests** build step, you can optionally specify the location of folders containing source code, relative to the project root folder. When you specify a folder in the **Folder path** box, the plugin adds the specified folder and its subfolders to the top of the MATLAB search path. You can specify more folders by clicking **Add Folder**. To generate a coverage report, MATLAB uses only the source code in the specified folders and their subfolders.
```
nbhoski, line 58: what happens if we leave this blank is kinda puzzle still as this statement `To generate a coverage report, MATLAB uses only the source code in the specified folders and their subfolders t ` is telling that if i select source it will be considered for coverage generation however it does not tell me if this is mandatory if you want to generate code coverage. (Which may not be the case because by default we consider the current workspace as source)

nbhoski (approved):
Looks good to me few minor comments feel free to drop them

CONFIGDOC.md
mw-hrastega: Thank you, Nikhil. I see your point. I changed the sentence to this:

If you specify folders, MATLAB uses only the source code in the specified folders and their subfolders to generate a coverage report.

PR #158, opened 2020-10-15
Handle apostrophe in file path value

abhargav-mw:
Bring back Bug Fixes from 2.1.4 Qual branch to dev_main


nbhoski (approved):
After the second run Job is passing so merge it.

PR #161, opened 2020-10-21
Add support for MATLAB in Global Tool Configuration

abhargav-mw:
This PR consists of the following,

-  **Feature:** Support for MATLAB in Global Tool Configuration.

- **Fixes:** Potential fix for issues in Windows machines in Azure Pipeline. Updated the dummy empty binaries of MATLAB in test resources with some executable code. This would avoid raising a prompt saying 'This app can't run on this PC' which probably caused the sporadic failures.


src/main/resources/com/mathworks/ci/MatlabInstallation/config.jelly
```diff
@@ -0,0 +1,9 @@
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
+    <f:entry title="Name: " field="name">
+        <f:textbox />
+    </f:entry>
+    <f:entry title="Matlab Root: " field="home">
```
mcafaro, line 6: Is this a UI label? If so, I think MATLAB should be capitalized and the label should be in sentence case:

Matlab Root -> MATLAB root

src/main/java/com/mathworks/ci/MatlabInstallation.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @Override
+    public MatlabInstallation forNode(Node node, TaskListener log) throws IOException, InterruptedException {
+        return new MatlabInstallation(getName(), translateFor(node, log), getProperties().toList());
+    }
+
+    @Override
+    public void buildEnvVars(EnvVars env) {
+        String pathToExecutable = getHome() + "/bin";
```
mcafaro, line 42: Should this always be "/bin"? Wouldn't it be "\bin" on Windows? Or does Jenkins take care of that automatically?

src/main/resources/com/mathworks/ci/MatlabInstallation/config.jelly
abhargav-mw: Done

src/main/java/com/mathworks/ci/MatlabInstallation.java
abhargav-mw: Yes, Jenkins (or windows) takes care of it. Currently, this class does not have access to build context hence extracting OS platform information is tricky.

mcafaro (approved):

mw-akumar (approved):

nbhoski (approved):
LGTM . however its  worth to see if we could throw an appropriate Exception when tools with matalb is used and there is no tool added in Jenkins. I assume currently it throws Hudson tools Exception stack trace which may not so informative.

abhargav-mw:
I'll explore better stack trace as part of next submissions.

PR #173, opened 2020-12-22
Add support for MATLAB tool installation in Freestyle Projects

abhargav-mw:


src/main/resources/config.properties
```diff
@@ -13,6 +13,7 @@ builder.matlab.automatictestoption.display.name = Automatic
 builder.matlab.customcommandoption.display.name = Custom
 Releaseinfo.matlab.version.not.found.error = Error finding MATLAB release for given MATLAB root. Verify MATLAB root path.
 matlab.not.found.error = Unable to launch MATLAB from the specified location. Verify the path to MATLAB root folder.
+matlab.not.found.error.for.node = Unable to launch MATLAB registered with the name %s on the node %s. Verify global tool configuration for the specified node.
```
mw-hrastega, line 16: How about shortening the first sentence this way (assuming the second sentence clarifies that it is a tool)?

Unable to launch MATLAB %s on the node %s.

src/main/resources/config.properties
```diff
@@ -13,6 +13,7 @@ builder.matlab.automatictestoption.display.name = Automatic
 builder.matlab.customcommandoption.display.name = Custom
 Releaseinfo.matlab.version.not.found.error = Error finding MATLAB release for given MATLAB root. Verify MATLAB root path.
```
mw-hrastega, line 14: This error message needs revision. How different is it from the message below it?

src/main/resources/config.properties
```diff
@@ -21,4 +22,11 @@ build.workspace.computer.not.found = Unable to access the computer for this buil
 matlab.command.build.step.name = runMATLABCommand
 matlab.tests.build.step.name = runMATLABTests
 matlab.command.step.display.name = Run MATLAB commands, scripts, or functions
-matlab.tests.step.display.name = Run MATLAB tests and generate artifacts
\ No newline at end of file
+matlab.tests.step.display.name = Run MATLAB tests and generate artifacts
+matlab.custom.location = Custom...
+matlab.added.to.path.from = Prepending MATLAB to PATH variable from this location: %s
```
mw-hrastega, line 27: to PATH --> to the PATH

src/main/resources/config.properties
```diff
@@ -21,4 +22,11 @@ build.workspace.computer.not.found = Unable to access the computer for this buil
...
+Axis.matlab.key = MATLAB
+Axis.use.matlab.warning = This project specifies MATLAB versions using the added 'MATLAB' axis as well as the 'Use MATLAB version' option. The values specified by 'Use MATLAB version' take precedence over the values specified by the 'MATLAB' axis.
+Axis.no.installed.matlab.error = Because the project does not have any MATLAB versions configured in Jenkins Global Tool Configuration, the plugin runs the default configuration.
+Use.matlab.version.axis.warning = This project specifies MATLAB version using the 'Use MATLAB version' option as well as the added 'MATLAB' axis. The values specified by 'Use MATLAB version' take precedence over the values specified by the 'MATLAB' axis.
```
mw-hrastega, line 31: version or versions?
When can there be only one version? When can there be multiple versions?

src/main/resources/config.properties
```diff
@@ -21,4 +22,11 @@ build.workspace.computer.not.found = Unable to access the computer for this buil
 matlab.command.build.step.name = runMATLABCommand
 matlab.tests.build.step.name = runMATLABTests
 matlab.command.step.display.name = Run MATLAB commands, scripts, or functions
-matlab.tests.step.display.name = Run MATLAB tests and generate artifacts
\ No newline at end of file
+matlab.tests.step.display.name = Run MATLAB tests and generate artifacts
+matlab.custom.location = Custom...
+matlab.added.to.path.from = Prepending MATLAB to PATH variable from this location: %s
+Axis.matlab.key = MATLAB
+Axis.use.matlab.warning = This project specifies MATLAB versions using the added 'MATLAB' axis as well as the 'Use MATLAB version' option. The values specified by 'Use MATLAB version' take precedence over the values specified by the 'MATLAB' axis.
```
mw-hrastega, line 29: Do you think the first sentence is necessary?

src/main/resources/config.properties
```diff
@@ -21,4 +22,11 @@ build.workspace.computer.not.found = Unable to access the computer for this buil
...
+Axis.no.installed.matlab.error = Because the project does not have any MATLAB versions configured in Jenkins Global Tool Configuration, the plugin runs the default configuration.
```
mw-hrastega, line 30: Reading the first clause, I think we can summarize it this way. Is either of these statements accurate?

-Because the project does not use any MATLAB tools, ...
-Because no MATLAB tools have been specified, ...

I also think it makes sense here to replace "runs" with "will run", because the plugin's action will definitely take place in the future.

src/main/resources/config.properties
mw-hrastega: I am saying this because the warning message might become quite verbose in the UI. It will be nice to see how it renders.

src/main/resources/com/mathworks/ci/UseMatlabVersionBuildWrapper/help.html
```diff
@@ -1,3 +1,9 @@
-<p>
-Specify the MATLAB version for this build.
-</p>
\ No newline at end of file
+<div>
+    <p>
+        Use this list to specify the MATLAB version for the build. The plugin prepends your specified MATLAB to the PATH environment variable and uses it to perform the build. <br>
+        The list contains the names of all MATLAB versions that have been registered as a tool in <u>Jenkins Global Tool Configuration</u>. Select one of the tools, or select <b>Custom…</b> if you want to specify a different version. When you select <b>Custom…</b>, you must enter the full path to the desired MATLAB root folder in the <b>MATLAB root</b> box.
```
mw-hrastega, line 4: The list contains --> The list includes
in <u>Jenkins Global Tool Configuration</u> --> in Jenkins

src/main/resources/com/mathworks/ci/UseMatlabVersionBuildWrapper/help.html
```diff
@@ -1,3 +1,9 @@
...
-</p>
\ No newline at end of file
+<div>
+    <p>
+        Use this list to specify the MATLAB version for the build. The plugin prepends your specified MATLAB to the PATH environment variable and uses it to perform the build. <br>
+        The list contains the names of all MATLAB versions that have been registered as a tool in <u>Jenkins Global Tool Configuration</u>. Select one of the tools, or select <b>Custom…</b> if you want to specify a different version. When you select <b>Custom…</b>, you must enter the full path to the desired MATLAB root folder in the <b>MATLAB root</b> box.
+    </p>
+    <p><br>
+        <b>Note:</b> For a Multi-configuration project that has 'MATLAB' axis added, 'Use MATLAB Version' value takes precedence over the values specified by the 'MATLAB' axis.
+    </p>
```
mw-hrastega, line 8: Not sure if this note is appropriate for free-style projects.

Some edits: 

<b>Note: </b>In multi-configuration projects, the values specified by <b>Use MATLAB version</b> take precedence over the values specified by the <b>MATLAB</b> axis.

mw-hrastega: What is the current help text for the "MATLAB root" box?

mw-hrastega (changes requested):
Requesting some edits and including some queries.

src/main/java/com/mathworks/ci/Utilities.java
```diff
@@ -23,4 +33,47 @@ public static String getCellArrayFrmList(List<String> listOfStr){
...
+            throws IOException, InterruptedException {
+        String name = env.get(Message.getValue("Axis.matlab.key"));
+
+        // If no MATLAB axis is set or if 'Use MATLAB version' is selected, return
+        if (name == null || name.isEmpty() || env.get("matlabroot") != null){
+            return;
+        }
+
+        String matlabExecutablePath = getNodeSpecificHome(name, cmp.getNode(), listener, env) + "/bin";
+        env.put("PATH+matlabroot", matlabExecutablePath);
```
mw-akumar, line 47: Is "PATH+matlabroot" the name of the environment variable?

src/main/java/com/mathworks/ci/Utilities.java
```diff
@@ -23,4 +33,47 @@ public static String getCellArrayFrmList(List<String> listOfStr){
...
+        // Specify which MATLAB was added to path.
+        listener.getLogger().println("\n" + String.format(Message.getValue("matlab.added.to.path.from"), matlabExecutablePath) + "\n");
+    }
+
+    public static String getNodeSpecificHome(String instName, Node node, TaskListener listener, EnvVars env)
+            throws IOException, InterruptedException {
+        MatlabInstallation inst = MatlabInstallation.getInstallation(instName);
```
mw-akumar, line 54: Perhaps a more readable version is to put the something like this at the top (early return)

if(inst == null) {
  return "";
}

src/main/java/com/mathworks/ci/Utilities.java
```diff
@@ -23,4 +33,47 @@ public static String getCellArrayFrmList(List<String> listOfStr){
...
+        if(inst != null) {
+            if (node != null) {
```
mw-akumar, line 56: If both node and env are not null, then value of inst computed in the first if, will be immediately overwritten by the next if. Consider refactoring.

mw-akumar (commented):
Some minor comments. Let me know if you want to discuss them in person.

src/main/resources/config.properties
abhargav-mw: This message is relevant for both 'MATLAB root' box in 'Use MATLAB Version' and the 'MATLAB root' box in 'Global Tool Configuration' page. This error message wouldn't need revision.

It differs from 'Unable to launch MATLAB... ' error message as this error only implies it's not able to find MATLAB release version and that the plugin wouldn't know if plugin supports a particular version of MATLAB. This is an unlikely edge case as MATLAB installation directory should have a Contents.m or VersionInfo.xml file.

src/main/resources/config.properties
abhargav-mw: If a user has selected 'Use MATLAB Version' there will be only one version. If that's not the case and a MALTAB axis is added there can be versions of MATLAB.

I propose we change 'values' --> 'value' in second part of the sentence. Both the warning should then look like (Notice in case of warning that will be displayed at axis I use 'versions' in first sentence),

Use.matlab.version.axis.warning = This project specifies MATLAB version using the 'Use MATLAB version' option as well as the added 'MATLAB' axis. The value specified by 'Use MATLAB version' take precedence over the values specified by the 'MATLAB' axis.

Axis.use.matlab.warning = This project specifies MATLAB versions using the added 'MATLAB' axis as well as the 'Use MATLAB version' option. The value specified by 'Use MATLAB version' take precedence over the values specified by the 'MATLAB' axis.

src/main/resources/config.properties
abhargav-mw: I think using versions instead of tools is a bit more accurate. MATLAB is a tool that has different versions that can be configured.

Using the second suggested clause, I propose the following revision:

Axis.no.installed.matlab.error = Because no MATLAB versions are configured in Jenkins Global Tool Configuration, the plugin will run the default matrix configuration.

I added '...matrix configuration.' to emphasize the difference in Global tool configuration and a Matrix configuration.

src/main/resources/com/mathworks/ci/UseMatlabVersionBuildWrapper/help.html
abhargav-mw: Agreed, I had added the 'Global Tool Configuration' as  a way of letting user know where MATLAB versions are configured. But perhaps this information is already exposed in out plugin documentation.

src/main/resources/com/mathworks/ci/UseMatlabVersionBuildWrapper/help.html
abhargav-mw: gbl_tool_helptext

This is what both the help text look like.

src/main/resources/config.properties
abhargav-mw: gbl_tool_all_warning

This is how it renders in UI

src/main/java/com/mathworks/ci/MatlabBuildWrapperContent.java
```diff
@@ -0,0 +1,30 @@
...
+import org.kohsuke.stapler.DataBoundConstructor;
+
+public class MatlabBuildWrapperContent {
+
+    private final String matlabInstName;
+    private final String matlabRootFolder;
+
+    @DataBoundConstructor
+    public MatlabBuildWrapperContent(String matlabInstName, String matlabRootFolder){
+        this.matlabInstName = matlabInstName;
```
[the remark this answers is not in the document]
abhargav-mw: Done.

src/main/java/com/mathworks/ci/MatlabInstallationAxis.java
```diff
@@ -0,0 +1,71 @@
...
+        // Add default configuration is values are null or not selected.
+        if (values == null || values.isEmpty()) {
+            values = new ArrayList<>(Arrays.asList("default"));
+        }
+        return values;
+    }
+
+    @Extension
+    public static class DescriptorImpl extends AxisDescriptor{
+        private final String useMATLABWarning = Message.getValue("Axis.use.matlab.warning");
```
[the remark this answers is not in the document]
abhargav-mw: Done.

src/main/java/com/mathworks/ci/Utilities.java
abhargav-mw: This is by design to get Installation for specific node and then specific environment. Updated the code to be more readable

src/main/java/com/mathworks/ci/Utilities.java
abhargav-mw: Done.

src/main/java/com/mathworks/ci/Utilities.java
abhargav-mw: Yes, this is by design. In Jenkins this would prepend matlabroot (represents  matlabExecutablePath) to PATH.

src/main/java/com/mathworks/ci/UseMatlabVersionBuildWrapper.java
```diff
@@ -136,17 +186,19 @@ public synchronized void setUp(Context context, Run<?, ?> build, FilePath worksp
         setEnv(initialEnvironment);
 
         FilePath matlabExecutablePath = new FilePath(launcher.getChannel(),
-                getLocalMatlab() + "/bin/" + getNodeSpecificExecutable(launcher));
+                getLocalMatlab(Computer.currentComputer(), listener) + "/bin/" + getNodeSpecificExecutable(launcher));
```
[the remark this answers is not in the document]
abhargav-mw: Updated the function name to be more readable.

src/main/java/com/mathworks/ci/UseMatlabVersionBuildWrapper.java
```diff
@@ -82,16 +108,40 @@ public String getDisplayName() {
...
+            arr.add(new MatlabInstallation(customLocation));
+            MatlabInstallation[] temp = new MatlabInstallation[arr.size()];
+            return arr.toArray(temp);
+        }
+
+        public String getCustomLocation() {
+            return customLocation;
+        }
+
+        public boolean getIsMatrix() {
```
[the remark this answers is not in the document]
abhargav-mw: Yes, this is used by jelly file to conditionally render UI elements. This function needs to be public.

src/main/java/com/mathworks/ci/UseMatlabVersionBuildWrapper.java
```diff
@@ -44,13 +45,41 @@ public String getMatlabRootFolder() {
...
+            throws IOException, InterruptedException {
+        return Utilities.getNodeSpecificHome(this.matlabInstName, cmp.getNode(), listener, env);
+    }
+
+    public String getMatlabInstName() {
+        /* For backward compatibility assign installation name to custom
+         * if matlabRootFolder is not null.
+         * */
+        if(this.matlabRootFolder!=null && !this.matlabRootFolder.isEmpty()){
+            this.matlabInstName = Message.getValue("matlab.custom.location");
```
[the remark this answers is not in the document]
abhargav-mw: This is a special case for backward compatibility. maltabInstallationName needs to be set to custom if maltabrootfolder has a value.

src/main/java/com/mathworks/ci/MatlabItemListener.java
```diff
@@ -0,0 +1,84 @@
...
+
+    void checkElementsSingle(Item item){
+        check((MatrixProject) item);
+    }
+
+    void check(MatrixProject _prj){
+        boolean checkForAxis = false;
+        boolean checkForBuildWrapper = false;
+        Collection<MatrixConfiguration> conf = _prj.getActiveConfigurations();
+        for(MatrixConfiguration _conf : conf){
```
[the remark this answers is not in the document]
abhargav-mw: Done.

src/main/java/com/mathworks/ci/MatlabItemListener.java
```diff
@@ -0,0 +1,84 @@
...
+    private static final Map<String, Boolean> prjCheckMATLABBuildWrapper = new HashMap<>();
+
+    @Override
+    public void onLoaded(){
+        checkElements(Jenkins.get().getItems());
+    }
+
+    @Override
+    public void onUpdated(Item item) {
+        if(!(item instanceof MatrixProject)){
```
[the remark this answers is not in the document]
abhargav-mw: Using instanceof is a sign of broken design if it is used to execute different behavior of subclasses than the superclass, as it signifies using superclass reference does not limit to super class functions. 

In this scenario, we use instanceof just as a way of knowing what Item was loaded or updated in Jenkins instance. Beyond checking the project type we do not use specific project properties or functions.

src/main/java/com/mathworks/ci/MatlabInstallationAxis.java
```diff
@@ -0,0 +1,71 @@
...
+
+        public boolean checkUseMATLABVersion(Object it) {
+            return MatlabItemListener.getMATLABBuildWrapperCheckForPrj(((MatrixProject) it).getFullName()) && !checkMatlabInstallationEmpty();
+        }
+
+        public MatlabInstallation[] getInstallations () {
+            return MatlabInstallation.getAll();
+        }
+
+        public String getUseMATLABWarning() {
```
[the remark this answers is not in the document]
abhargav-mw: Yes, these are used by jelly files to render UI.

src/main/java/com/mathworks/ci/MatlabInstallation.java
```diff
@@ -43,6 +59,23 @@ public void buildEnvVars(EnvVars env) {
...
+        return Jenkins.get().getDescriptorByType(DescriptorImpl.class).getInstallations();
+    }
+
+    public static boolean isEmpty() {
+        return getAll().length == 0;
+    }
+
+    public static MatlabInstallation getInstallation(String name) {
+        for (MatlabInstallation _inst : getAll()) {
+            if (!name.isEmpty() && name.equals(_inst.getName())) {
```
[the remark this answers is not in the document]
abhargav-mw: Thank you for pointing this out, removed the redundant check. :)

src/main/resources/com/mathworks/ci/UseMatlabVersionBuildWrapper/help.html
abhargav-mw: Applied the suggested edits.

src/main/resources/config.properties
abhargav-mw: Applied the suggested edits.

src/main/resources/com/mathworks/ci/UseMatlabVersionBuildWrapper/help.html
```diff
@@ -1,3 +1,9 @@
...
-Specify the MATLAB version for this build.
-</p>
\ No newline at end of file
+<div>
+    <p>
+        Use this list to specify the MATLAB version for the build. The plugin prepends your specified MATLAB to the PATH environment variable and uses it to perform the build. <br>
+        The list includes the names of all MATLAB versions that have been registered as a tool in <u>Jenkins</u>. Select one of the tools, or select <b>Custom…</b> if you want to specify a different version. When you select <b>Custom…</b>, you must enter the full path to the desired MATLAB root folder in the <b>MATLAB root</b> box.
+    </p>
+    <p><br>
+        <b>Note:</b> In multi-configuration projects, the values specified by <b>Use MATLAB Version</b> value takes precedence over the values specified by the <b>MATLAB</b> axis.
```
mw-hrastega, line 7: Please make this change:

by <b>Use MATLAB Version</b> value takes ---> 
by <b>Use MATLAB version</b> takes

Details:
-To match the UI, "v" in version should not be capitalized.
-"value" after Use MATLAB version doesn't make sense.

src/main/resources/com/mathworks/ci/UseMatlabVersionBuildWrapper/help.html
mw-hrastega: A few comments on the help text for MATLAB root:

1) Now that we have shipped 20b, it makes sense to change R2019a to R2020b.

2) Change Mac to macOS. 

3) Rewrite the second sentence to be in active voice:
The plugin prepends the folder to the PATH environment variable and uses the specified MATLAB version to perform the build.

3) Rewrite the rest of the paragraph as follows:
You must not leave this box blank if no MATLAB folder exists on the path. If you leave it blank, the plugin attempts to use the topmost MATLAB folder on the system path.

5) The word "Example" alone looks weird. Probably you can replace it with something like this:
Here are examples of the root folder path on different platforms:

src/main/resources/config.properties
mw-hrastega: Thanks for clarifying. I just think the error message needs to be reworded. It sounds awkward. How about something like this?

Unable to find the MATLAB version corresponding to the specified path. Verify the path to MATLAB root folder.

* find or locate

src/main/resources/config.properties
```diff
@@ -21,4 +22,11 @@ build.workspace.computer.not.found = Unable to access the computer for this buil
...
 matlab.command.step.display.name = Run MATLAB commands, scripts, or functions
-matlab.tests.step.display.name = Run MATLAB tests and generate artifacts
\ No newline at end of file
+matlab.tests.step.display.name = Run MATLAB tests and generate artifacts
+matlab.custom.location = Custom...
+matlab.added.to.path.from = Prepending MATLAB to the PATH variable from this location: %s
+Axis.matlab.key = MATLAB
+Axis.use.matlab.warning = This project specifies MATLAB versions using the added 'MATLAB' axis as well as the 'Use MATLAB version' option. The value specified by 'Use MATLAB version' take precedence over the values specified by the 'MATLAB' axis.
+Axis.no.installed.matlab.error = Because no MATLAB versions are configured in Jenkins Global Tool Configuration, the plugin will run the default matrix configuration.
+Use.matlab.version.axis.warning = This project specifies MATLAB version using the 'Use MATLAB version' option as well as the added 'MATLAB' axis. The value specified by 'Use MATLAB version' take precedence over the values specified by the 'MATLAB' axis.
```
mw-hrastega, line 31: 1) Here in the first sentence version should be changed to versions. This is because multiple versions are specified using two options. 
2) In the second sentence, take should be takes.

src/main/resources/config.properties
```diff
@@ -21,4 +22,11 @@ build.workspace.computer.not.found = Unable to access the computer for this buil
 matlab.command.build.step.name = runMATLABCommand
 matlab.tests.build.step.name = runMATLABTests
 matlab.command.step.display.name = Run MATLAB commands, scripts, or functions
-matlab.tests.step.display.name = Run MATLAB tests and generate artifacts
\ No newline at end of file
+matlab.tests.step.display.name = Run MATLAB tests and generate artifacts
+matlab.custom.location = Custom...
+matlab.added.to.path.from = Prepending MATLAB to the PATH variable from this location: %s
+Axis.matlab.key = MATLAB
+Axis.use.matlab.warning = This project specifies MATLAB versions using the added 'MATLAB' axis as well as the 'Use MATLAB version' option. The value specified by 'Use MATLAB version' take precedence over the values specified by the 'MATLAB' axis.
```
mw-hrastega, line 29: take ---> takes

src/main/resources/config.properties
```diff
@@ -21,4 +22,11 @@ build.workspace.computer.not.found = Unable to access the computer for this buil
...
+Axis.no.installed.matlab.error = Because no MATLAB versions are configured in Jenkins Global Tool Configuration, the plugin will run the default matrix configuration.
```
mw-hrastega, line 30: Because no MATLAB versions are configured in Jenkins Global Tool Configuration,

Let's replace "are" with "have been" or use something like this:

Because no MATLAB versions exist under Jenkins Global Tool Configuration,

* If you want to go with the first sentence, is configure the right verb? How about register?

Also, I prefer under Jenkins to in Jenkins.

mw-hrastega (changes requested):
Looks better. Suggesting some additional tweaks.

src/main/resources/com/mathworks/ci/UseMatlabVersionBuildWrapper/help.html
abhargav-mw: Done.

src/main/resources/config.properties
abhargav-mw: Done.

src/main/resources/config.properties
abhargav-mw: Done.

src/main/resources/config.properties
abhargav-mw: Done. Chose the second suggestion.

src/main/java/com/mathworks/ci/MatlabItemListener.java
```diff
@@ -0,0 +1,91 @@
...
+                check((MatrixProject) item);
+            }
+        }
+    }
+
+    private void checkSingleItem(Item item){
+        check((MatrixProject) item);
+    }
+
+    private void check(MatrixProject _prj) {
```
mcafaro, line 55: Why do some variable names start with underscores? This doesn't seem to be a consistent pattern across the code.

src/main/java/com/mathworks/ci/MatlabInstallationAxis.java
```diff
@@ -0,0 +1,71 @@
...
+        public String getDisplayName() {
+            return Message.getValue("Axis.matlab.key");
+        }
+
+        @Override
+        public boolean isInstantiable() {
+            return !checkMatlabInstallationEmpty();
+        }
+
+        public boolean checkUseMATLABVersion(Object it) {
```
mcafaro, line 51: `MATLAB` -> `Matlab`

There's a few more around the rest of the code as well. You might do a find-replace (watch out for user-facing text though, where it should be MATLAB).

src/main/resources/config.properties
```diff
@@ -13,6 +13,7 @@ builder.matlab.automatictestoption.display.name = Automatic
 builder.matlab.customcommandoption.display.name = Custom
 Releaseinfo.matlab.version.not.found.error = Error finding MATLAB release for given MATLAB root. Verify MATLAB root path.
 matlab.not.found.error = Unable to launch MATLAB from the specified location. Verify the path to MATLAB root folder.
+matlab.not.found.error.for.node = Unable to launch MATLAB %s on the node %s. Verify global tool configuration for the specified node.
```
mcafaro, line 16: May consider placing single quotes around the substituted strings to make it clear to the user that they are configured values:

```suggestion
matlab.not.found.error.for.node = Unable to launch MATLAB '%s' on the node '%s'. Verify global tool configuration for the specified node.
```

mcafaro (approved):
Looks good to me. Just some style notes.

mw-akumar (approved):

src/main/java/com/mathworks/ci/MatlabInstallationAxis.java
```diff
@@ -0,0 +1,71 @@
...
+            return MatlabItemListener.getMATLABBuildWrapperCheckForPrj(((MatrixProject) it).getFullName()) && !checkMatlabInstallationEmpty();
```
nbhoski, line 52: checkMatlabInstallationEmpty() does not tell what its doing especially when its negating the results. 
may be isMatlabInstallationEmpty() reads better.

src/main/java/com/mathworks/ci/MatlabInstallationAxis.java
```diff
@@ -0,0 +1,71 @@
...
+        // Add default configuration is values are null or not selected.
+        if (values == null || values.isEmpty()) {
+            values = new ArrayList<>(Arrays.asList("default"));
+        }
+        return values;
+    }
+
+    @Extension
+    public static class DescriptorImpl extends AxisDescriptor{
+        private final String useMatlabWarning = Message.getValue("Axis.use.matlab.warning");
```
nbhoski, line 38: would separate final string make any difference if the message is pulled from a static property file.

abhargav-mw: Not on performance but it does add to readability and emphasize that it's a constant value in classes context.

https://softwareengineering.stackexchange.com/questions/250804/why-to-declare-a-string-as-final-and-then-use-it

src/main/java/com/mathworks/ci/MatlabItemListener.java
```diff
@@ -0,0 +1,91 @@
...
+import jenkins.model.Jenkins;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+@Extension
+public final class MatlabItemListener extends ItemListener {
+    private static final Map<String, Boolean> prjCheckMATLABAxis = new HashMap<>();
```
nbhoski, line 26: +1 on @mcafaro  's comments for using MATLAB for customer facing UI and Matlab othrwise

src/main/java/com/mathworks/ci/MatlabItemListener.java
nbhoski: +1 on using consistent variable patterns across code

src/main/java/com/mathworks/ci/MatlabItemListener.java
```diff
@@ -0,0 +1,91 @@
...
+    private void check(MatrixProject _prj) {
+        checkForAxis(_prj);
+        checkForBuildWrapper(_prj);
+    }
+
+    private void checkForAxis(MatrixProject _prj) {
+        boolean checkForAxis = false;
+        Collection<MatrixConfiguration> conf = _prj.getActiveConfigurations();
+        for(MatrixConfiguration _conf : conf){
+            String a = _conf.getCombination().get(Message.getValue("Axis.matlab.key"));
```
nbhoski, line 64: String 'a' is not intuitive about what it is used for.

src/main/java/com/mathworks/ci/MatlabItemListener.java
```diff
@@ -0,0 +1,91 @@
...
+            }
+        }
+        prjCheckMATLABAxis.put(_prj.getFullName(), checkForAxis);
+    }
+
+    private void checkForBuildWrapper(MatrixProject _prj) {
+        boolean checkForBuildWrapper = false;
+        for(Object _bWrapper : _prj.getBuildWrappersList().toArray()) {
+            if(_bWrapper instanceof UseMatlabVersionBuildWrapper){
+                checkForBuildWrapper = ((UseMatlabVersionBuildWrapper) _bWrapper).getMatlabInstallationName() != null || ((UseMatlabVersionBuildWrapper) _bWrapper).getMatlabRootFolder() != null;
```
nbhoski, line 77: How does it behave when we  click on the use MATLAB checkbox but does not provide any values in the text box for matlabroo.

nbhoski (commented):
Some minor changes .

src/main/java/com/mathworks/ci/MatlabItemListener.java
abhargav-mw: Good catch! For purpose of this class, this condition works fine even if matlabroot is empty or null. Actually here it wouldn't matter if matlabroot textbox is empty as by design we always assign matlabInstallationName a value if 'Use MATLAB version' is selected, even in case of 'Custom...' 

So perhaps I need to remove the second part of the OR condition as it is never reached and is redundant.

src/main/java/com/mathworks/ci/MatlabItemListener.java
abhargav-mw: Done. Removed the underscores.

src/main/java/com/mathworks/ci/MatlabInstallationAxis.java
abhargav-mw: Done.

src/main/resources/config.properties
abhargav-mw: Done.

src/main/java/com/mathworks/ci/MatlabInstallationAxis.java
abhargav-mw: Done.

src/main/java/com/mathworks/ci/MatlabItemListener.java
abhargav-mw: Done

src/main/java/com/mathworks/ci/MatlabItemListener.java
abhargav-mw: Done.

mw-hrastega (approved):

src/main/java/com/mathworks/ci/MatlabInstallationAxis.java
nbhoski: I guess again the benefits will boil down to reduce the total line number in the code. I would choose one over the other just to reduce the number of extra lines of code

abhargav-mw: Done.

src/test/java/com/mathworks/ci/MatlabInstallationTest.java
```diff
@@ -21,10 +28,40 @@
...
+            } catch (URISyntaxException e) {
+                e.printStackTrace();
+            }
+        } else {
+            FileSeperator = "\\";
+            url = classLoader.getResource("com/mathworks/ci/win/bin/matlab.bat");
+            matlabExecutorAbsolutePath = new File(url.toURI()).getAbsolutePath();
+        }
+    }
+
```
nbhoski, line 64: Not sure why line 65 onwards is not showing as newly added block. I hope its newly added class all together

src/test/java/com/mathworks/ci/MatlabInstallationTest.java
```diff
@@ -85,4 +122,90 @@ public void verifyInstallationInPipeline() throws Exception {
         jenkins.assertBuildStatusSuccess(build);
         jenkins.assertLogContains("versioninfo", build);
     }
+    /*
```
nbhoski, line 125: at line 82 in this file we are using  matlabInstDescriptor.get installations(); and in MatlabInstallation class at line 76 we also have similar static method which returns the same we do you think these methods are be redundant in usage ? 
for eg : if we already have a static method we could use that across the code whenevr we need array of Installations
matlabInstDescriptor.getInstallations()  --> MatlabInstallations.getAll() just to keep the implementation consistent

nbhoski:
I see some indentation issue here

https://github.com/mathworks/jenkins-matlab-plugin/blob/9df77690679c94db39b029b539e757ae53c96f6e/src/main/java/com/mathworks/ci/UseMatlabVersionBuildWrapper.java#L37

nbhoski (approved):
Current implementation loks good to me with some minor comments.

src/test/java/com/mathworks/ci/MatlabInstallationTest.java
abhargav-mw: It is a pre-existing class, I added newer tests to it.

src/test/java/com/mathworks/ci/MatlabInstallationTest.java
abhargav-mw: Done.

PR #189, opened 2021-04-28
2.5.0 qualification

abhargav-mw:
**Added Feature:**

- Raise an Exception in pipeline execution if MATLAB exits with a non-zero code.


src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
```diff
@@ -43,10 +43,15 @@ public Void run() throws Exception {
         
         workspace.mkdirs();
         
-        int res = execMatlabCommand(workspace, launcher, listener, env);
+        int exitCode = execMatlabCommand(workspace, launcher, listener, env);
 
-        getContext().setResult((res == 0) ? Result.SUCCESS : Result.FAILURE);
-        
+        if(exitCode == 0){
```
mw-akumar, line 48: Perhaps it's a matter of style, but having the error condition in an if condition and returning feels more readable. Also, it allows you to add more code below the error condition in case you want to do more things before returning. 

if(exitCode !=0 {
 // throw an exception if return code is non-zero
        stop(new MatlabExecutionException(exitCode));
        return null;
}

getContext().setResult(Result.SUCCESS);
return null;

src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
```diff
@@ -43,10 +43,15 @@ public Void run() throws Exception {
...
+            getContext().setResult(Result.SUCCESS);
+            return null;
+        }
+
+        // throw an exception if return code is non-zero
+        stop(new MatlabExecutionException(exitCode));
```
mw-akumar, line 54: Does a the result not have to be set here?

Vahila:
Qualification checks made in Windows, Linux and Mac:

- Exception raising and the exception message in DSL and scripted pipelines
- Checked for backwards compatibility

Vahila (approved):
Qualification checks made in Windows, Linux and Mac:

- Exception raising and the exception message in DSL and scripted pipelines
- Checked for backwards compatibility

src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
abhargav-mw: Nope. This will raise exception to pipeline execution level and automatically set build result to fail if exception is not handled by user in pipeline script.

PR #190, opened 2021-05-11
Run tests in parallel

BrianCian:
Adding support for RunInParallel

### DIFF
--- a/src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
+++ b/src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
@@ -50,6 +50,8 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
     private transient boolean stmResultsChkBx;
     private transient boolean modelCoverageChkBx;
     private transient boolean pdfReportChkBx;
+
+    private boolean runTestsInParallel;
     
     private Artifact tapArtifact = new NullArtifact();
     private Artifact junitArtifact = new NullArtifact();
@@ -99,7 +101,12 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
     public void setPdfReportArtifact(PdfArtifact pdfReportArtifact) {
         this.pdfReportArtifact = pdfReportArtifact;
     }
-    
+
+    @DataBoundSetter
+    public void setRunTestsInParallel(boolean runTestsInParallel) {
+        this.runTestsInParallel = runTestsInParallel;
+    }
+
     @DataBoundSetter
     public void setSelectByTag(SelectByTag selectByTag) {
         this.selectByTag = selectByTag;
@@ -159,7 +166,11 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
     public Artifact getPdfReportArtifact() {
         return this.pdfReportArtifact;
     }
-    
+
+    public boolean getRunTestsInParallel() {
+        return this.runTestsInParallel;
+    }
+
     public String getPdfReportFilePath() {
         return this.getPdfReportArtifact().getFilePath();
     }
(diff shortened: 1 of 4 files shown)


src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -355,6 +366,9 @@ private String getInputArguments() {
             getSelectByTag().addTagToInputArgs(inputArgsList);
         }
 
+        // Add Run tests in parallel choice
+        inputArgsList.add("'RunInParallel'" + "," + getRunTestsInParallel());
```
nbhoski, line 370: should we pass this to genscript if its not checked ?

src/main/java/com/mathworks/ci/RunMatlabTestsStep.java
```diff
@@ -155,8 +165,10 @@ private String getInputArgs() {
         inputArgs.add("'Test'");
 
         args.forEach((key, val) -> {
-            if(key.equals("SourceFolder") || key.equals("SelectByFolder") && val != null){
+            if(key.equals("SourceFolder") || key.equals("SelectByFolder") && val != null) {
                 inputArgs.add("'" + key + "'" + "," + val);
+            }else if (key.equals("RunInParallel") && val != null){
+                inputArgs.add("'" + key + "'" + "," + Boolean.parseBoolean(val));
```
nbhoski, line 171: when this will be null ? since its boolean

src/main/java/com/mathworks/ci/RunMatlabTestsStep.java
```diff
@@ -174,11 +186,12 @@ private String getInputArgs() {
         args.put("CoberturaCodeCoverage", getCodeCoverageCobertura());
         args.put("CoberturaModelCoverage", getModelCoverageCobertura());
         args.put("SelectByTag", getSelectByTag());
+        args.put("RunInParallel", getRunTestsInParallel());
```
nbhoski, line 189: does it directly store boolean value in <String,string> type map ? did you consider String.valueOf() instead.

src/main/java/com/mathworks/ci/RunMatlabTestsStep.java
nbhoski: Aah i see its string . But does it make sense to pass string value here ? i can still pass 'ABC'  which will return false . I dont have strong flag on this though

src/main/java/com/mathworks/ci/RunMatlabTestsStep.java
nbhoski: Same as above

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/config.jelly
```diff
@@ -4,10 +4,16 @@
 <f:block>
     <br></br>
     <f:optionalProperty field="sourceFolder" title="Source folder" />
+    <br></br>
 </f:block>
```
nbhoski, line 8: Corresponding Help files are still misiing

src/main/resources/com/mathworks/ci/RunMatlabTestsStep/config.jelly
```diff
@@ -38,4 +38,8 @@
 		<f:textbox/>
 	</f:entry>
```
nbhoski, line 39: Corresponding help files are still missing

nbhoski (changes requested):

nbhoski:
Also , These changes require doc update. make sure we have corresponding changes either in same PR or separate PR and is approved.

BrianCian:
Ran into an issue with getting this to work with Artifacts. Also was having trouble getting help to work the 'checkbox'. Closing for now as I will be working on another project, but will come back to this at a later date.

PR #194, opened 2021-06-23
Asynch build tasks

nbhoski:
After this change both Run test and Run command tasks for freestyle and pipeline will support parallel run on the same host.
Matrix builds will now run job asynchronously unlike earlier behavior where builds were synchronized.

### DIFF
--- a/src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
+++ b/src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
@@ -24,7 +24,7 @@ public class MatlabCommandStepExecution extends SynchronousNonBlockingStepExecut
     }
 
     private String getCommand() {
-        return this.env == null ? this.command : this.env.expand(this.command );
+        return this.command;
     }
     
     private void setEnv(EnvVars env) {
@@ -37,7 +37,6 @@ public class MatlabCommandStepExecution extends SynchronousNonBlockingStepExecut
         final FilePath workspace = getContext().get(FilePath.class);
         final TaskListener listener = getContext().get(TaskListener.class);
         final EnvVars env = getContext().get(EnvVars.class);
-        setEnv(env);
         
         //Make sure the Workspace exists before run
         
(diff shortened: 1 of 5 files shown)


src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
```diff
@@ -91,12 +90,13 @@ private void createMatlabScriptByName(FilePath uniqeTmpFolderPath, String unique
         // Create a new command runner script in the temp folder.
         final FilePath matlabCommandFile =
                 new FilePath(uniqeTmpFolderPath, uniqueScriptName + ".m");
+        final String cmd = getContext().get(EnvVars.class).expand(getCommand());
```
mcafaro, line 93: With this change, is the class-level field `env` used anymore? If not, can it be removed along with `setEnv`?

src/main/java/com/mathworks/ci/RunMatlabCommandBuilder.java
```diff
@@ -96,21 +96,21 @@ public void perform(@Nonnull Run<?, ?> build, @Nonnull FilePath workspace,
             @Nonnull Launcher launcher, @Nonnull TaskListener listener)
             throws InterruptedException, IOException {
 
-        // Set the environment variable specific to the this build
-        setEnv(build.getEnvironment(listener));
+        // Get the environment variable specific to the this build
+        final EnvVars env = build.getEnvironment(listener);
```
mcafaro, line 100: Is the class-level field `env` used anymore? If not, can it be removed along with `setEnv`?

src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
```diff
@@ -59,7 +58,7 @@ public void stop(Throwable cause) throws Exception {
         getContext().onFailure(cause);
     }
     
-    private synchronized int execMatlabCommand(FilePath workspace, Launcher launcher,
+    private int execMatlabCommand(FilePath workspace, Launcher launcher,
```
mcafaro, line 61: Why did this previously need to be `synchronized`?

src/main/java/com/mathworks/ci/RunMatlabCommandBuilder.java
nbhoski: Done.

src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
nbhoski: Done.

src/main/java/com/mathworks/ci/MatlabCommandStepExecution.java
nbhoski: Right, none of these required it if we had not set environment in perform method.

mcafaro (approved):

Issue #202, opened 2021-11-02
scriptgen package reported in coverage report, artificially lowering coverage percentage


mcafaro:
In version 2.7.0, the scriptgen package is now listed in coverage reports, which artificially lowers coverage percentage.

PR #203, opened 2021-11-18
Fixed github isuue 202

nbhoski:
Fixes isuue #202 
In order to fix the issue following is implemented 
* changed the runner script to generate actual scriptgen script
* Runner script now does following 
    * Generate actual script based on user inputs 
    * keep the script in randomly generated temp folder under system temp folder 
    * Adds the above folder in path to access the script from workspace.
    * Deletes all scriptgen packages and related `.m files. and exits 
 * The actual call includes 
    * the runner script call 
    * deleting the runner script itself from .matlab folder
    * Running the actual script from temp folder 
    * Deleting the temp folder from system temp folder location including the script


nbhoski:
Linux image has some issue in allotment will re run.

src/main/java/com/mathworks/ci/MatlabBuilderConstants.java
```diff
@@ -41,5 +41,10 @@
     static final String TEST_RUNNER_SCRIPT = "testScript = genscript(${PARAMS});\n" + "\n"
             + "disp('Running MATLAB script with content:');\n"
             + "disp(strtrim(testScript.writeToText()));\n"
-            + "fprintf('___________________________________\\n\\n');\n" + "run(testScript);\n" + "";
+            + "runner_file = testScript.writeToFile([tempname(tempname),'.m']);\n"
```
mcafaro, line 44: Why is `tempname` being passed to `tempname`? If you're writing the script to a temp directory, can't the script just have a fixed, readable, name?

src/main/java/com/mathworks/ci/MatlabBuilderConstants.java
```diff
@@ -41,5 +41,10 @@
...
+            + "[tmpdir,runner_script,ext] = fileparts(runner_file);"
```
mcafaro, line 45: MATLAB variable names are typically camelCase

src/main/java/com/mathworks/ci/MatlabBuilderConstants.java
```diff
@@ -41,5 +41,10 @@
     static final String TEST_RUNNER_SCRIPT = "testScript = genscript(${PARAMS});\n" + "\n"
             + "disp('Running MATLAB script with content:');\n"
             + "disp(strtrim(testScript.writeToText()));\n"
```
mcafaro, line 43:
```suggestion
            + "disp(testScript.Contents);\n"
```

src/main/java/com/mathworks/ci/MatlabRunTestsStepExecution.java
```diff
@@ -68,12 +68,16 @@ private int execMatlabCommand(FilePath workspace, Launcher launcher,
                     "addpath('" + genScriptLocation.getRemote().replaceAll("'", "''") + "'); ";
             final String matlabScriptName = getValidMatlabFileName(genScriptLocation.getBaseName());
 
-            ProcStarter matlabLauncher = getProcessToRunMatlabCommand(workspace, launcher, listener,
-                    envVars, cmdPrefix + matlabScriptName, uniqueTmpFldrName);
+            ProcStarter matlabLauncher =
+                    getProcessToRunMatlabCommand(workspace, launcher, listener, envVars,
+                            cmdPrefix + matlabScriptName + ",delete .matlab/"
+                                    + genScriptLocation.getBaseName() + "/" + matlabScriptName
+                                    + ".m,run(runner_script),rmdir(tmpdir,'s')",
```
mcafaro, line 75: Need to be careful here. Looks like if `genScriptLocation.getBaseName()` or `matlabScriptName` has a space, this will fail. You may know they won't at the moment, but it's safer to just use the function syntax form of `delete`.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -301,7 +301,8 @@ private int execMatlabCommand(FilePath workspace, Launcher launcher,
     public String constructCommandForTest(FilePath scriptPath) {
         final String matlabScriptName = getValidMatlabFileName(scriptPath.getBaseName());
         final String runCommand = "addpath('" + scriptPath.getRemote().replaceAll("'", "''")
-                + "'); " + matlabScriptName;
+                + "'); " + matlabScriptName + ",delete .matlab/" + scriptPath.getBaseName() + "/"
+                + matlabScriptName + ".m,run(runner_script),rmdir(tmpdir,'s')";
```
mcafaro, line 305: Same comment as above, use the function form of `delete`.

src/main/java/com/mathworks/ci/MatlabBuilderConstants.java
```diff
@@ -41,5 +41,10 @@
...
-            + "fprintf('___________________________________\\n\\n');\n" + "run(testScript);\n" + "";
+            + "runner_file = testScript.writeToFile([tempname(tempname),'.m']);\n"
+            + "[tmpdir,runner_script,ext] = fileparts(runner_file);"
+            + "addpath(tmpdir);\n"
+            + "rmdir('.matlab/tmp_f/+scriptgen', 's');\n"
+            + "delete .matlab/tmp_f/genscript.m;\n"
```
mcafaro, line 48: Use function form of `delete`.

src/main/java/com/mathworks/ci/MatlabBuild.java
```diff
@@ -99,8 +99,9 @@ default void prepareTmpFldr(FilePath tmpFldr, String runnerScript) throws IOExce
         zipFileLocation.unzip(tmpFldr);
     }
     
-    default String getRunnerScript(String script, String params) {
+    default String getRunnerScript(String script, String params, String uniqueTmpFldrName) {
         script = script.replace("${PARAMS}", params);
+        script = script.replaceAll("tmp_f", uniqueTmpFldrName);
```
mcafaro, line 104: Why two different approaches to placeholder strings?
```suggestion
        script = script.replace("${PARAMS}", params);
        script = script.replaceAll("${TMPDIR}", uniqueTmpFldrName);

```

mcafaro (commented):
Tests?

src/main/java/com/mathworks/ci/MatlabBuilderConstants.java
nbhoski: yes, but since we are adding the temp in Path there is chance of readable static name collision in case of Matrix runs etc.

src/main/java/com/mathworks/ci/MatlabBuilderConstants.java
nbhoski: Done .

src/main/java/com/mathworks/ci/MatlabBuilderConstants.java
nbhoski: Done.

src/main/java/com/mathworks/ci/MatlabRunTestsStepExecution.java
nbhoski: These uses 8 bit UUID so i suspect it ever gets space in them.

src/main/java/com/mathworks/ci/MatlabBuild.java
nbhoski: tried this approach already however the occurrence $ is more than once i guess Java API .replaceAll throws an exception in its subsequent occurrence

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
nbhoski: does delete remove Directory ?

src/main/java/com/mathworks/ci/MatlabRunTestsStepExecution.java
nbhoski: oh ok got the point , are you hinting at use delete() instead delete <file> ?

src/main/java/com/mathworks/ci/MatlabBuilderConstants.java
nbhoski: Done

src/main/java/com/mathworks/ci/MatlabRunTestsStepExecution.java
nbhoski: Done.

src/main/java/com/mathworks/ci/MatlabBuild.java
mcafaro: Looks like `replaceAll` takes a regular expression, not just a raw string value. $ has meaning in regexp. You should be able to escape it with backslash. Note { also has special meaning in regexp, so you'll need to escape that as well.

src/main/java/com/mathworks/ci/MatlabBuilderConstants.java
mcafaro: Matrix runs don't run in the same MATLAB session though, right? Whatever you're adding to the path in one session is not going to effect what's going on in another session unless you savepath.

nbhoski: its different sessions correct. But how do you think would it behave when 
fld1/runner.m and fld2/runner.m got added at same/ different time in path and one session tries to call runner.m ? will it still be tightly coupled to the session ? I prefer name being unique even if we have chance of name collision either by user or by our code itself even once in million.

src/main/java/com/mathworks/ci/MatlabBuild.java
nbhoski: Done

src/main/java/com/mathworks/ci/MatlabBuilderConstants.java
```diff
@@ -40,6 +40,11 @@
     // MATLAB runner script
     static final String TEST_RUNNER_SCRIPT = "testScript = genscript(${PARAMS});\n" + "\n"
             + "disp('Running MATLAB script with content:');\n"
-            + "disp(strtrim(testScript.writeToText()));\n"
-            + "fprintf('___________________________________\\n\\n');\n" + "run(testScript);\n" + "";
+            + "disp(strtrim(testScript.Contents));\n"
+            + "runnerFile = testScript.writeToFile([tempname(tempname),'.m']);\n"
```
acampbel, line 44: fullfile(tempname, 'runnerScript.m')

src/main/java/com/mathworks/ci/MatlabBuild.java
```diff
@@ -99,8 +99,9 @@ default void prepareTmpFldr(FilePath tmpFldr, String runnerScript) throws IOExce
         zipFileLocation.unzip(tmpFldr);
```
acampbel, line 99: Something to think about for the future, would be pretty clean if we could swing only copying the zip file into the .matlab folder and then unzipping it into the containers local temp folder. This way, there never is any "m-files" that show up in the workspace last all and could get mistakenly used in code coverage.

src/main/java/com/mathworks/ci/MatlabRunTestsStepExecution.java
```diff
@@ -68,12 +68,16 @@ private int execMatlabCommand(FilePath workspace, Launcher launcher,
                     "addpath('" + genScriptLocation.getRemote().replaceAll("'", "''") + "'); ";
             final String matlabScriptName = getValidMatlabFileName(genScriptLocation.getBaseName());
 
-            ProcStarter matlabLauncher = getProcessToRunMatlabCommand(workspace, launcher, listener,
-                    envVars, cmdPrefix + matlabScriptName, uniqueTmpFldrName);
+            ProcStarter matlabLauncher =
+                    getProcessToRunMatlabCommand(workspace, launcher, listener, envVars,
+                            cmdPrefix + matlabScriptName + ",delete('.matlab/"
+                                    + genScriptLocation.getBaseName() + "/" + matlabScriptName
+                                    + ".m'),run(runnerScript),rmdir(tmpDir,'s')",
```
acampbel, line 75: I thought that last week we ran into a problem with the run function adding a problematic cd. Why isn't that a problem here?

src/main/java/com/mathworks/ci/MatlabRunTestsStepExecution.java
```diff
@@ -68,12 +68,16 @@ private int execMatlabCommand(FilePath workspace, Launcher launcher,
...
+                            uniqueTmpFldrName);
             
             // prepare temp folder by coping genscript package and writing runner script.
             prepareTmpFldr(genScriptLocation,
-                    getRunnerScript(MatlabBuilderConstants.TEST_RUNNER_SCRIPT, envVars.expand(getCommandArgs())));
+                    getRunnerScript(MatlabBuilderConstants.TEST_RUNNER_SCRIPT, envVars.expand(getCommandArgs()),uniqueTmpFldrName));
```
acampbel, line 80: Should we rename uniqueTmpFolderName? To me this sounds like it is in the /tmp folder, not the .matlab folder.

acampbel (commented):
+ 1 on brainstorming to find a way to write az test that would fail if this were to regress

src/main/java/com/mathworks/ci/MatlabBuilderConstants.java
nbhoski: Done.

src/main/java/com/mathworks/ci/MatlabBuild.java
nbhoski: Sure , Tracking it with issue #205

src/main/java/com/mathworks/ci/MatlabRunTestsStepExecution.java
nbhoski: Remember now, ts because earlier the temp file was directly written into /tmp instead /tmp/random_folder. hence it was trying to CD some  folder in /tmp and running into issue there. for safer side i have removed  run () anyways

src/main/java/com/mathworks/ci/MatlabRunTestsStepExecution.java
nbhoski: Done. changed to uniqueMatlabResourceFldr

src/test/java/com/mathworks/ci/RunMatlabTestsBuilderTest.java
```diff
@@ -500,4 +500,28 @@ public void verifyMATLABfolderGenerated() throws Exception {
         File matlabRunner = new File(build.getWorkspace() + File.separator + ".matlab");
         Assert.assertTrue(matlabRunner.exists());
     }
+    
+    /*
+     * Test to verify if runner script is deleted from .matlb folder during MATLAB call
```
acampbel, line 505: typo ".matlb"

acampbel (approved):
Thanks, these tests are a good start, although ideally we get to a point where our unit tests are able to qualify the intended behavior more directly.

abhargav-mw (approved):

src/test/java/com/mathworks/ci/RunMatlabTestsBuilderTest.java
nbhoski: Done.

PR #204, opened 2021-11-22
Fix for MATLAB root path displayed in Jenkins build logs.

abhargav-mw:
This fixes the unmatched slashes in the MATLAB root path

### DIFF
--- a/src/main/java/com/mathworks/ci/Utilities.java
+++ b/src/main/java/com/mathworks/ci/Utilities.java
@@ -45,8 +45,9 @@ public class Utilities {
         }
 
         String matlabExecutablePath = getNodeSpecificHome(name,
-               cmp.getNode(), listener, env) + "/bin";
+               cmp.getNode(), listener, env) + ((Boolean.TRUE.equals(cmp.isUnix()))?"/bin" : "\\bin");
         env.put("PATH+matlabroot", matlabExecutablePath);
+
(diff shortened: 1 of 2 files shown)


mw-hrastega:
Hi, Anmol. The changes look good to me.

src/main/java/com/mathworks/ci/Utilities.java
```diff
@@ -45,8 +45,9 @@ public static void addMatlabToEnvPathFrmAxis(Computer cmp, TaskListener listener
         }
 
         String matlabExecutablePath = getNodeSpecificHome(name,
-               cmp.getNode(), listener, env) + "/bin";
+               cmp.getNode(), listener, env) + ((Boolean.TRUE.equals(cmp.isUnix()))?"/bin" : "\\bin");
```
nbhoski, line 48: Why do we need Boolean.TRUE.equal ? does't just cmp.isUnix() ? a : b works

nbhoski (approved):
LGTM , with some minor comment. re run azure tc once i push my changes in.

src/main/java/com/mathworks/ci/Utilities.java
abhargav-mw: isUnix() has @ CheckForNull annotation and may return null pointer exception, This is a fail-safe

acampbel (approved):

PR #219, opened 2022-08-01
remove runnerfile from .matlab folder

dolonm-mw:
Moving the runnerfile from .matlab folder to system temporary folder.

### DIFF
--- a/src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
+++ b/src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
@@ -301,8 +301,8 @@ public class RunMatlabTestsBuilder extends Builder implements SimpleBuildStep, M
     public String constructCommandForTest(FilePath scriptPath) {
         final String matlabScriptName = getValidMatlabFileName(scriptPath.getBaseName());
         final String runCommand = "addpath('" + scriptPath.getRemote().replaceAll("'", "''")
-                + "'); " + matlabScriptName + ",delete('.matlab/" + scriptPath.getBaseName() + "/"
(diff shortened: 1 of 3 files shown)


src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -301,8 +301,8 @@ private int execMatlabCommand(FilePath workspace, Launcher launcher,
     public String constructCommandForTest(FilePath scriptPath) {
         final String matlabScriptName = getValidMatlabFileName(scriptPath.getBaseName());
         final String runCommand = "addpath('" + scriptPath.getRemote().replaceAll("'", "''")
-                + "'); " + matlabScriptName + ",delete('.matlab/" + scriptPath.getBaseName() + "/"
-                + matlabScriptName + ".m'),runnerScript,rmdir(tmpDir,'s')";
+                + "');tmpDir=tempname;mkdir(tmpDir), ,movefile('.matlab/" + scriptPath.getBaseName() + "/"
```
nbhoski, line 304: Why move file here, Cant we keep the runner script as is and just call it here with simple script name

nbhoski (changes requested):
I don't think we are making much difference here by removing delete with movefile.

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
nbhoski: @dolonm-mw can you pls look into this

dolonm-mw: On it!

PR #231, opened 2022-12-07
Run build

davidbuzinski:
Initial "Run MATLAB Build" integration for buildtool
**For Freestyle project:**
- Run MATLAB Build option with Tasks parameter (String) that takes a space-separated list of buildtool tasks

**For DSL:**
- runMATLABBuild step with tasks parameter (String) that takes a space-separated list of buildtool tasks

### DIFF
--- a/src/main/java/com/mathworks/ci/MatlabBuildStepExecution.java
+++ b/src/main/java/com/mathworks/ci/MatlabBuildStepExecution.java
@@ -1,0 +1,102 @@
+package com.mathworks.ci;
+
+import java.io.IOException;
+import org.jenkinsci.plugins.workflow.steps.StepContext;
+import org.jenkinsci.plugins.workflow.steps.SynchronousNonBlockingStepExecution;
+import hudson.EnvVars;
+import hudson.FilePath;
+import hudson.Launcher;
+import hudson.Launcher.ProcStarter;
+import hudson.model.Result;
+import hudson.model.TaskListener;
+
+public class MatlabBuildStepExecution extends SynchronousNonBlockingStepExecution<Void> implements MatlabBuild {
+    
+    private static final long serialVersionUID = 6704588180717665100L;
+    
+    private String tasks;
+
+    public MatlabBuildStepExecution(StepContext context, String tasks) {
+        super(context);
+        this.tasks = tasks;
+    }
+
+    private String getTasks() {
+        return this.tasks;
+    }
+
+    @Override
+    public Void run() throws Exception {
+        final Launcher launcher = getContext().get(Launcher.class);
+        final FilePath workspace = getContext().get(FilePath.class);
+        final TaskListener listener = getContext().get(TaskListener.class);
+        final EnvVars env = getContext().get(EnvVars.class);
+        
+        //Make sure the Workspace exists before run
+        
+        workspace.mkdirs();
+        
+        int exitCode = execMatlabCommand(workspace, launcher, listener, env);
+
+        if(exitCode != 0){
+            // throw an exception if return code is non-zero
+            stop(new MatlabExecutionException(exitCode));
+        }
+
+        getContext().setResult(Result.SUCCESS);
+        return null;
+    }
+
+    @Override
+    public void stop(Throwable cause) throws Exception {
+        getContext().onFailure(cause);
+    }
+    
+    private int execMatlabCommand(FilePath workspace, Launcher launcher,
+            TaskListener listener, EnvVars envVars) throws IOException, InterruptedException {
+        final String uniqueTmpFldrName = getUniqueNameForRunnerFile();  
+        final String uniqueBuildFile =
+                "build_" + getUniqueNameForRunnerFile().replaceAll("-", "_");
+        final FilePath uniqueTmpFolderPath =
+                getFilePathForUniqueFolder(launcher, uniqueTmpFldrName, workspace);
+
+        // Create MATLAB script
+        createMatlabScriptByName(uniqueTmpFolderPath, uniqueBuildFile, workspace, listener);
+        ProcStarter matlabLauncher;
+
+        try {
+            matlabLauncher = getProcessToRunMatlabCommand(workspace, launcher, listener, envVars,
+                    "cd('"+ uniqueTmpFolderPath.getRemote().replaceAll("'", "''") +"'); "+ uniqueBuildFile, uniqueTmpFldrName);
+            listener.getLogger()
+                    .println("#################### Starting command output ####################");
+            return matlabLauncher.pwd(workspace).join();
+
+        } finally {
+            // Cleanup the tmp directory
+            if (uniqueTmpFolderPath.exists()) {
+                uniqueTmpFolderPath.deleteRecursive();
+            }
+        }
+    }
+    
+    private void createMatlabScriptByName(FilePath uniqeTmpFolderPath, String uniqueScriptName, FilePath workspace, TaskListener listener) throws IOException, InterruptedException {
+
+        // Create a new command runner script in the temp folder.
+        final FilePath matlabBuildFile =
+                new FilePath(uniqeTmpFolderPath, uniqueScriptName + ".m");
+        final String tasks = getContext().get(EnvVars.class).expand(getTasks());
+        String cmd = "buildtool";
+
+        if (!tasks.trim().isEmpty()) {
+            cmd += " " + tasks;
+        }
+        final String matlabBuildFileContent =
+                "cd '" + workspace.getRemote().replaceAll("'", "''") + "';\n" + cmd;
+
+        // Display the commands on console output for users reference
+        listener.getLogger()
+                .println("Generating MATLAB script with content:\n" + cmd + "\n");
+
+        matlabBuildFile.write(matlabBuildFileContent, "UTF-8");
+    }
+}
--- a/src/main/java/com/mathworks/ci/MatlabBuildStepExecution.java
+++ b/src/main/java/com/mathworks/ci/MatlabBuildStepExecution.java
@@ -1,5 +1,10 @@
 package com.mathworks.ci;
 
+/**
+ * Copyright 2022 The MathWorks, Inc.
+ *  
+ */
+
 import java.io.IOException;
 import org.jenkinsci.plugins.workflow.steps.StepContext;
 import org.jenkinsci.plugins.workflow.steps.SynchronousNonBlockingStepExecution;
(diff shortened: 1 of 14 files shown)


src/main/java/com/mathworks/ci/MatlabBuildStepExecution.java
```diff
@@ -0,0 +1,107 @@
...
+import hudson.EnvVars;
+import hudson.FilePath;
+import hudson.Launcher;
+import hudson.Launcher.ProcStarter;
+import hudson.model.Result;
+import hudson.model.TaskListener;
+
+public class MatlabBuildStepExecution extends SynchronousNonBlockingStepExecution<Void> implements MatlabBuild {
+    
+    private static final long serialVersionUID = 6704588180717665100L;
```
davidbuzinski, line 20: Not sure where to get this from. It might need updated. @nikhilbhoski

src/main/java/com/mathworks/ci/MatlabBuildStepExecution.java
```diff
@@ -0,0 +1,107 @@
...
+        super(context);
+        this.tasks = tasks;
+    }
+
+    private String getTasks() {
+        return this.tasks;
+    }
+
+    @Override
+    public Void run() throws Exception {
```
davidbuzinski, line 34: This is almost the same in all of our Matlab(Action)StepExecutions. Might be a way to reuse or abstract in the future

src/main/java/com/mathworks/ci/MatlabBuildStepExecution.java
```diff
@@ -0,0 +1,107 @@
...
+        getContext().setResult(Result.SUCCESS);
+        return null;
+    }
+
+    @Override
+    public void stop(Throwable cause) throws Exception {
+        getContext().onFailure(cause);
+    }
+    
+    private int execMatlabCommand(FilePath workspace, Launcher launcher,
```
davidbuzinski, line 60: This is almost identical in several files. Might be a way to abstract or reuse in the future

src/main/java/com/mathworks/ci/MatlabBuildStepExecution.java
```diff
@@ -0,0 +1,107 @@
...
+    
+    private void createMatlabScriptByName(FilePath uniqeTmpFolderPath, String uniqueScriptName, FilePath workspace, TaskListener listener) throws IOException, InterruptedException {
+
+        // Create a new command runner script in the temp folder.
+        final FilePath matlabBuildFile =
+                new FilePath(uniqeTmpFolderPath, uniqueScriptName + ".m");
+        final String tasks = getContext().get(EnvVars.class).expand(getTasks());
+        String cmd = "buildtool";
+
+        if (!tasks.trim().isEmpty()) {
```
davidbuzinski, line 95: I don't love that the logic to make the MATLAB command is mixed with the logic to create script in the `createMatlabScriptByName` method

src/main/java/com/mathworks/ci/RunMatlabBuildBuilder.java
```diff
@@ -0,0 +1,152 @@
...
+
+
+        buildResult = execMatlabCommand(workspace, launcher, listener, env);
+
+        if (buildResult != 0) {
+            build.setResult(Result.FAILURE);
+        }
+    }
+
+    private int execMatlabCommand(FilePath workspace, Launcher launcher,
```
davidbuzinski, line 93: Lots of duplication in the `execMatlabCommand` and `createMatlabScriptByName` methods in all 6 of the Builder and StepExecution files. Something to consider going forward

src/test/java/com/mathworks/ci/RunMatlabBuildBuilderTest.java
```diff
@@ -0,0 +1,335 @@
...
+        scriptBuilder.setTasks("$TASKS");
+        project.getBuildersList().add(scriptBuilder);
+        FreeStyleBuild build = project.scheduleBuild2(0).get();
+        jenkins.assertLogContains("buildtool compile", build);
+    }
+    
+    /*
+     * Test to verify if appropriate MATLAB runner file is copied in workspace.
+     * 
+     * NOTE: This test assumes there is no MATLAB installed and is not on System Path.
```
davidbuzinski, line 243: Would love to remove this requirement in the future (in all other tests as well). It took me a while to figure out why tests were failing when I first tried to build.

src/test/java/com/mathworks/ci/RunMatlabBuildBuilderTest.java
```diff
@@ -0,0 +1,335 @@
...
+        scriptBuilder.setTasks("");
+        project.getBuildersList().add(scriptBuilder);
+        FreeStyleBuild build = project.scheduleBuild2(0).get();
+        jenkins.assertLogContains("MatlabNotFoundError", build);
+    }
+    
+	/*
+	 * Test to verify if Matrix build fails when MATLAB is not available.
+     * 
+     * NOTE: This test assumes there is no MATLAB installed and is not on System Path.
```
davidbuzinski, line 272: Same here. I'd love to remove this requirement

src/test/java/com/mathworks/ci/RunMatlabBuildBuilderTester.java
```diff
@@ -0,0 +1,117 @@
...
+import hudson.Launcher.ProcStarter;
+import hudson.model.AbstractProject;
+import hudson.model.Result;
+import hudson.model.Run;
+import hudson.model.TaskListener;
+import hudson.tasks.BuildStepDescriptor;
+import hudson.tasks.Builder;
+import net.sf.json.JSONObject;
+
+public class RunMatlabBuildBuilderTester extends RunMatlabBuildBuilder {
```
davidbuzinski, line 26: There is a Tester for each of the types of MATLAB tasks that are almost all the same. We might be able to consolidate here going forward as well (Mock MATLAB executor for all of them?)

src/test/java/com/mathworks/ci/RunMatlabBuildStepTest.java
```diff
@@ -0,0 +1,158 @@
...
+        j.assertBuildStatusSuccess(build);
+    }
+
+    /*
+     * Verify appropriate task is invoked as in pipeline script
+     */
+    @Test
+    public void verifyTasksSameAsScript() throws Exception {
+        project.setDefinition(
+                new CpsFlowDefinition("node { runMATLABBuild(tasks: 'compile') }", true));
```
davidbuzinski, line 96: A lot of these tests (and the builder tests) feel more like integration tests, and several of them take a while to run. Maybe some abstractions can make the code more directly testable

src/main/java/com/mathworks/ci/MatlabBuildStepExecution.java
nbhoski: Most IDE's provide auto generation of this UID like eclipse and this should be unique as during serelizations classes are identified with these ID's for now you could safly use `1L` since other values are been auto generated

src/test/java/com/mathworks/ci/RunMatlabBuildBuilderTest.java
nbhoski: the rationale behind  tests is  considering the environment of jenkins.  where it would run without actual MATLAB instances. Hence whatever we could test which is closer to actual would be good idea

src/test/java/com/mathworks/ci/RunMatlabBuildBuilderTest.java
nbhoski: Same as above

src/test/java/com/mathworks/ci/RunMatlabBuildBuilderTester.java
nbhoski: +1

src/main/java/com/mathworks/ci/RunMatlabBuildBuilder.java
nbhoski: +1

src/main/resources/com/mathworks/ci/RunMatlabBuildBuilder/help-tasks.html
```diff
@@ -0,0 +1,11 @@
+<div>
+    Insert a space-separated list of tasks to run in the <b>Tasks</b> box. If not specified, the action runs the default tasks in buildfile.m as well as all the tasks on which they depend.<br>
+    <b>Example: </b>test<br>
+    <b>Example: </b>compile test
+    <br>&nbsp;</br>
+    <b>Note:</b>
+    <ul>
+        <li>MATLAB exits with exit code 0 if the tasks run without error. Otherwise, MATLAB terminates with a nonzero exit code, which causes the workflow to fail.</li>
```
nbhoski, line 8: Does it make sense to replace workflow with build.

src/test/java/com/mathworks/ci/RunMatlabBuildBuilderTest.java
nbhoski: I also think we could restrict the test cases related to RunMatlabBuilderBuild Class here instead of re running duplicates.

src/test/java/com/mathworks/ci/RunMatlabBuildStepTest.java
nbhoski: alternately as i commented earlier we could just restrict test related to RunMatlabBuildstep class.

src/main/java/com/mathworks/ci/MatlabBuildStepExecution.java
nbhoski: Yeah we could declare in MatlabBuild interface

src/test/java/com/mathworks/ci/RunMatlabBuildBuilderTest.java
davidbuzinski: I am having trouble understanding which duplicate tests to remove. `verifyMATLABrunnerFileGenerated` and `verifyMatrixBuildFails`?

src/main/resources/com/mathworks/ci/RunMatlabBuildBuilder/help-tasks.html
davidbuzinski: Yes good catch

src/main/java/com/mathworks/ci/MatlabBuildStepExecution.java
davidbuzinski: Ok, I was able to install Eclipse and generate one. For some reason VS Code wasn't generating one for me, even with a Java language extension.

PR #236, opened 2023-02-24
update to make tasks options for runMATLABBuild step

davidbuzinski:
This change makes tasks optional in the `runMATLABBuild` step. It will no longer accept tasks as a positional argument. I tried overloading the constructor to see if I could accept an optional positional argument, while also taking the named `tasks` option, but this doesn't seem compatible with Jenkins.

The following will no longer error:
```
node {
    runMATLABBuild()
}
```

The following is now the only way to specify tasks:
```
node {
    runMATLABBuild(tasks: 'compile test')
}
```

This no longer works:
```
node {
    runMATLABBuild 'compile test'
}
```


mcafaro (approved):
:shipit:

PR #250, opened 2023-07-19
Binary update

sameagen-MW:
Updated plugin to use the `run-matlab-command` binary and added support for passing startup options through to MATLAB to maintain parity with cloud CI/CD platforms.

I ended up adding `startupOptions` as an additional argument on `getProcessToRunMatlabCommand` because I preferred the mental model of handling them separately, but I also think that the options could be baked into the command as well if others prefer that implementation.

Besides the small tweaks to support startup options, the primary change here is the artifacts that the pom downloads, and then some of the logic within `MatlabBuild`. Since there's now three option instead of two, I used a call to `uname` to differentiate between MacOS and Linux.


src/main/resources/com/mathworks/ci/RunMatlabBuildBuilder/config.jelly
```diff
@@ -5,4 +5,6 @@
 	        <f:textbox/>
 	  </f:entry> 
 
-</j:jelly>
\ No newline at end of file
+      <f:optionalProperty field="startupOptions" title="Startup Options" />
```
mw-hrastega, line 8: "Startup Options" or "MATLAB Startup Options"?

src/main/resources/com/mathworks/ci/RunMatlabBuildStep/config.jelly
```diff
@@ -1,8 +1,12 @@
...
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
     
-	  <f:entry field="tasks">
+	  <f:entry field="tasks" title="Tasks: ">
 	        <f:textbox/>
 	  </f:entry> 
 
-</j:jelly>
\ No newline at end of file
+      <f:entry field="startupOptions" title="Startup Options: ">
```
mw-hrastega, line 8: "Startup Options" or "MATLAB Startup Options"?

src/main/resources/com/mathworks/ci/RunMatlabCommandStep/help-startupOptions.html
```diff
@@ -0,0 +1,3 @@
+<div>
+ A space-separated list of MATLAB Startup Options.
```
mw-hrastega, line 2: s   o

src/main/resources/com/mathworks/ci/RunMatlabTestsStep/help-startupOptions.html
```diff
@@ -0,0 +1,3 @@
+<div>
+ A space-separated list of MATLAB Startup Options.
```
mw-hrastega, line 2: s o

src/main/resources/com/mathworks/ci/StartupOptions/help-startupOptions.html
```diff
@@ -0,0 +1,3 @@
+<div>
+ A space-separated list of MATLAB Startup Options.
```
mw-hrastega, line 2: s o

src/main/resources/com/mathworks/ci/RunMatlabBuildStep/config.jelly
mw-hrastega: Same question for anywhere else in this submission. I can see both "MATLAB Startup Options" and "Startup Options".

src/main/resources/com/mathworks/ci/RunMatlabBuildBuilder/config.jelly
sameagen-MW: Hmm, which do you prefer?

mw-hrastega: Probably MATLAB startup options to clarify that these are not plugin/Jenkins server startup options?

mw-hrastega: Again, it depends on the context provided. I am leaning more towards "Startup options" if the context is clear. That's how we named the argument in the cloud-based platforms.

mw-hrastega: Also, when would it be possible to see how the step UIs look including the new feature? Or is this submission not relevant to the UI?

sameagen-MW: The UIs have already been updated if you pull down the repo and check out the BinaryUpdate branch. I can also send over screenshots if that's easier.

sameagen-MW: I agree with you, I think that "Startup Options" is more readable. "MATLAB Startup Options" is much more daunting visually and draws attention that it perhaps doesn't need to.

src/main/resources/com/mathworks/ci/RunMatlabBuildStep/config.jelly
mw-hrastega, line 8: I wonder why we use a colon in only a subset of titles. I can see both title="Startup Options: " and title="Startup Options".

sameagen-MW: Good point, I think that this was a mistake on my part. All of the builders use "Startup Options," and I tried to stick with the precedent in each of the files. For the *Steps, I thought I was going off of the precedent of "RunMatlabTestsStep" but I think I misrembered and changed "startupOptions: " into "Startup Options:". I'll get that fixed.

src/main/resources/com/mathworks/ci/RunMatlabTestsBuilder/config.jelly
```diff
@@ -84,7 +84,7 @@
 </f:block>
 </f:section>
 
-<f:section title="MATLAB Startup Options">
+<f:section title="Speficy MATLAB Startup Options">
```
mw-hrastega: Specify

sameagen-MW:
I don't know why but I can't respond direclty to your comment Houman. Those changes seem reasonable to me though

sameagen-MW:
# Updated
image
image

src/main/java/com/mathworks/ci/RunMatlabCommandBuilder.java
```diff
@@ -114,10 +124,11 @@ private int execMatlabCommand(FilePath workspace, Launcher launcher,
         // Create MATLAB script
         createMatlabScriptByName(uniqeTmpFolderPath, uniqueCommandFile, workspace, listener, envVars);
         ProcStarter matlabLauncher;
+        String options = getStartupOptions() == null ? "" : getStartupOptions().getStartupOptions();
```
nbhoski, line 127: getStartupOptions().getStartupOptions() reads redundant may be getStartupOptions().getAllOptions() or getOptions()

src/main/java/com/mathworks/ci/RunMatlabTestsBuilder.java
```diff
@@ -342,12 +348,13 @@ private int execMatlabCommand(FilePath workspace, Launcher launcher,
 
         final String uniqueTmpFldrName = getUniqueNameForRunnerFile();
         ProcStarter matlabLauncher;
+        String options = getStartupOptions() == null ? "" : getStartupOptions().getStartupOptions();
```
nbhoski, line 351: Same as Above

src/main/resources/com/mathworks/ci/StartupOptions/help-startupOptions.html
```diff
@@ -0,0 +1,3 @@
+<div>
+ A space-separated list of MATLAB startup options.
```
nbhoski, line 2: Should we have link to showup the actual option list ? How one would learn about these options

src/main/java/com/mathworks/ci/RunMatlabBuildBuilder.java
```diff
@@ -108,10 +119,10 @@ private int execMatlabCommand(FilePath workspace, Launcher launcher,
         // Create MATLAB script
         createMatlabScriptByName(uniqeTmpFolderPath, uniqueBuildFile, workspace, listener, envVars);
         ProcStarter matlabLauncher;
-
+        String options = getStartupOptions() == null ? "" : getStartupOptions().getStartupOptions();
         try {
             matlabLauncher = getProcessToRunMatlabCommand(workspace, launcher, listener, envVars,
-                    "cd('"+ uniqeTmpFolderPath.getRemote().replaceAll("'", "''") +"');"+ uniqueBuildFile, uniqueTmpFldrName);
+                    "cd('"+ uniqeTmpFolderPath.getRemote().replaceAll("'", "''") +"');"+ uniqueBuildFile, options, uniqueTmpFldrName);
```
nbhoski, line 125: Do we pass "" here if startupOptions is NULL  like uniqueFile,,  curious to see the log output when no options set

nbhoski (changes requested):
I am not sure what exactly the startup options affect the workflow and should we include it. since we have included it it also then requires Doc update with examples. @mw-hrastega do you agree?

src/main/java/com/mathworks/ci/MatlabBuild.java
```diff
@@ -29,28 +30,50 @@ public interface MatlabBuild {
...
-                    targetWorkspace);
+            copyFileInWorkspace(binaryName, runnerName, targetWorkspace);
         } else {
-            final String runnerScriptName = uniqueName + "\\run_matlab_command.bat";
-            launcher = launcher.decorateByPrefix("cmd.exe", "/C");
+            targetWorkspace = new FilePath(launcher.getChannel(),
+                workspace.getRemote() + "\\" + MatlabBuilderConstants.TEMP_MATLAB_FOLDER_NAME);
+
+            final String runnerName = uniqueName + "\\run-matlab-command.exe";
+            // launcher = launcher.decorateByPrefix("cmd.exe", "/C");
```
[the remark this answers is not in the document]
sameagen-MW: Yep. We're just launching the `run-matlab-command(.exe)` program directly as a new process. That program then uses the standard Go "os/exec" library to run the final command (either using `matlab-batch` or `matlab` directly).

https://javadoc.jenkins-ci.org/hudson/Launcher.html

https://github.mathworks.com/development/run-matlab-command/blob/main/src/main/go/pkg/internal/command/executor.go

src/main/java/com/mathworks/ci/MatlabBuild.java
```diff
@@ -29,28 +30,50 @@ public interface MatlabBuild {
...
             matlabLauncher = launcher.launch().envs(envVars);
-            matlabLauncher.cmds(MatlabBuilderConstants.TEMP_MATLAB_FOLDER_NAME + "\\" + runnerScriptName, "\"" + matlabCommand + "\"")
+            matlabLauncher.cmds(targetWorkspace.toString() + "\\" + runnerName, "\"" + matlabCommand + "\"", startupOpts)
                     .stdout(listener);
+
             // Copy runner.bat for Windows platform in workspace.
-            copyFileInWorkspace(MatlabBuilderConstants.BAT_RUNNER_SCRIPT, runnerScriptName,
+            copyFileInWorkspace("win64\\run-matlab-command.exe", runnerName,
```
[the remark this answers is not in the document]
sameagen-MW: Hmm, I think that this is a good suggestion. The part that makes me hesitate here is that we're already having to hardcode in the program name on lines 54, 57, and 69. I don't really like keeping that information in two separate places, since it means if the program name were to change it would be easy for a user to just change the constant name and miss the fact that there are also hardcoded instances. 

I don't have a strong preference either way, it just feels a little bit odd to me to have a property in another file define one instance and a hardcoded string define another. There's also a slight readability sacrifice when future contributors have to cross reference between the file they're working on and the constants file but that's not a real issue in the long run. 

Also I realize I left a line of code commented out, I'll update the pull request to remove that (and also probably condense lines 54 and 57 to be one line after the if block).

What are your thoughts @davidbuzinski?

sameagen-MW:
> I am not sure what exactly the startup options affect the workflow and should we include it. since we have included it it also then requires Doc update with examples. @mw-hrastega do you agree?

Just to double check that I understand what you mean, you would have a preference for only updating the Jenkins plugin to use `run-matlab-command` and not adding support for passing startup options through to MATLAB?

nbhoski:
Hi @sameagen-MW  coud you just confirm if we consider all CR as resolved ?  I got confused seeing few unresolved comments

src/main/resources/com/mathworks/ci/StartupOptions/help-startupOptions.html
sameagen-MW: Good question, I wonder if @mw-hrastega has a preference here?

sameagen-MW:
> Hi @sameagen-MW coud you just confirm if we consider all CR as resolved ? I got confused seeing few unresolved comments

Yeah totally, I've gone and resolved all of the conversations that I think are totally finished.

src/main/java/com/mathworks/ci/MatlabBuild.java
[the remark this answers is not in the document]
davidbuzinski: I don't have a strong opinion unless it's being used all over the code. I think it's basically just in 1 or 2 places right? I'm fine wither way

src/main/resources/com/mathworks/ci/StartupOptions/help-startupOptions.html
mw-hrastega: @sameagen-MW: I think we should be consistent with how other options are represented in the software. If others do not include links, we cannot provide a link here.

src/main/java/com/mathworks/ci/MatlabBuild.java
```diff
@@ -29,28 +30,47 @@ public interface MatlabBuild {
...
+            ByteArrayOutputStream kernelStream = new ByteArrayOutputStream();
+            launcher.launch()
+                .cmds("uname")
+                .masks(true)
+                .stdout(kernelStream)
+                .join();
+
+            String binaryName;
+            String runnerName = uniqueName + "/run-matlab-command";
+            if (kernelStream.toString("UTF-8").contains("Linux")) {
```
davidbuzinski, line 52: This is probably fine, but I'm more used to special casing to mac and defaulting to Linux rather than visa versa. Not a huge deal though

src/main/resources/com/mathworks/ci/StartupOptions/help-startupOptions.html
sameagen-MW: That makes sense to me!

src/main/java/com/mathworks/ci/RunMatlabCommandBuilder.java
```diff
@@ -114,10 +124,11 @@ private int execMatlabCommand(FilePath workspace, Launcher launcher,
         // Create MATLAB script
         createMatlabScriptByName(uniqeTmpFolderPath, uniqueCommandFile, workspace, listener, envVars);
         ProcStarter matlabLauncher;
+        String options = getStartupOptions() == null ? "" : getStartupOptions().getOptions();
 
         try {
             matlabLauncher = getProcessToRunMatlabCommand(workspace, launcher, listener, envVars,
-                    "cd('"+ uniqeTmpFolderPath.getRemote().replaceAll("'", "''") +"');"+ uniqueCommandFile, uniqueTmpFldrName);
+                    "cd('"+ uniqeTmpFolderPath.getRemote().replaceAll("'", "''") +"');"+ uniqueCommandFile, options, uniqueTmpFldrName);
```
davidbuzinski, line 131: just realized that uniqe (unique) is spelled wrong here lol. Not sure if this mistake is all over or how hard it is to fix

davidbuzinski (approved):

nbhoski (approved):
Ship it!

PR #251, opened 2023-09-11
2.10.1 snapshot qualification

nbhoski:
Binary Updates 

Updated plugin to use the run-matlab-command binary and added support for passing startup options through to MATLAB to maintain parity with cloud CI/CD platforms.

I ended up adding startupOptions as an additional argument on getProcessToRunMatlabCommand because I preferred the mental model of handling them separately, but I also think that the options could be baked into the command as well if others prefer that implementation.

Besides the small tweaks to support startup options, the primary change here is the artifacts that the pom downloads, and then some of the logic within MatlabBuild. Since there's now three option instead of two, I used a call to uname to differentiate between MacOS and Linux.


Vahila (approved):
Qualification checks made in Windows, Linux:

1. Verified that startup options work with all Jenkins projects.
2. Regression testing.
3. Backwards compatibility.

PR #252, opened 2023-09-27
Fix bug where ClassLoader was unable to find run-matlab-command.exe f…

sameagen-MW:
…or certain versions of Jenkins.

Pull request is WIP, verified change fixes issue for recent versions of Jenkins but need to double check that it doesn't break things on older versions.


nbhoski (approved):
LGTM

sameagen-MW:
Changed `RunMatlabCommandStep`'s constructor to only take in one argument and moved `startupOptions` into a `DataBoundSetter`. Also, I've verified that the change in path has not broken anything in older versions of Jenkins.

davidbuzinski (approved):

PR #256, opened 2023-10-09
Rewrite how startup options are parsed, so that they can be split

sameagen-MW:
Windows is tricky! I changed the `ProcLauncher` in MatlabBuild.java to take in an ordered set of arguments so that we can split `startupOpts` on spaces. This resolves the issue with CMD treating strings differently from bash. Let me know if you have any flags on the changes


davidbuzinski (approved):
:shipit:

nbhoski (approved):
LGTM

PR #269, opened 2023-11-09
Addressed  Jenkins security issues

nbhoski:
We have Jenkins issue opened with some security vulnerabilities  as per the suggestions this PR contains the fix. 
* ReproductionStep
Add a dangerous XML called VersionInfo.xml into the Jenkins controller file system using an "Execute Shell” build step.
Payload:
echo '
<!DOCTYPE foo [ <!ENTITY xxe SYSTEM "file://<CustomPath>"> ]>
<MathWorks_version_info>
<release>
&xxe;
</release>
</MathWorks_version_info>' > VersionInfo.xml
Add an “Archive the artifacts” post build step
**/*.xml
Build the Job
Go into the "configure job"
Check “Use MATLAB version" under the "Build Environment" section.
Fill the "MATLAB root" with ./work/jobs/<JobName>/builds/<BuildNumber>/archive/
OR
Navigate to:
<JenkinsInstanceURL>/descriptorByName/com.mathworks.ci.UseMatlabVersionBuildWrapper/checkMatlabRootFolder?matlabRootFolder=./work/jobs/<JobName>/builds/<BuildNumber>/archive/

### DIFF
--- a/src/main/java/com/mathworks/ci/MatlabBuilder.java
+++ b/src/main/java/com/mathworks/ci/MatlabBuilder.java
@@ -8,6 +8,7 @@ package com.mathworks.ci;
  * nikhil.bhoski@mathworks.in Date : 28/03/2018 (Initial draft)
  */
 
+import hudson.model.Item;
 import hudson.security.Permission;
 import java.io.File;
 import java.io.IOException;
@@ -20,6 +21,7 @@ import java.util.function.Function;
 import javax.annotation.Nonnull;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang.ArrayUtils;
+import org.kohsuke.stapler.AncestorInPath;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.DataBoundSetter;
 import org.kohsuke.stapler.QueryParameter;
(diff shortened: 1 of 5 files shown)


src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -148,8 +149,9 @@ public DescriptorExtensionList<TestRunTypeList, Descriptor<TestRunTypeList>> get
          * descriptor class.
          */
 
-
+        @POST
         public FormValidation doCheckMatlabRoot(@QueryParameter String matlabRoot) {
+            Jenkins.get().checkPermission(Jenkins.ADMINISTER);
```
yaroslavafenkin, line 154: This breaks the UI for users with Job/Configure permission:
image

nbhoski: I am using Permission.CREATE now would that be the safe one to check ?

yaroslavafenkin: See https://www.jenkins.io/doc/developer/security/form-validation/#checking-permissions
I see that you're not using CREATE, but CONFIGURE. It has to be checked on the correct context. Please see the docs above, second code example there. `@AncestorInPath` is what you're interested in.

src/main/java/com/mathworks/ci/MatlabBuilder.java
```diff
@@ -148,8 +150,9 @@ public DescriptorExtensionList<TestRunTypeList, Descriptor<TestRunTypeList>> get
          * descriptor class.
          */
 
-
+        @POST
         public FormValidation doCheckMatlabRoot(@QueryParameter String matlabRoot) {
+            Jenkins.get().checkPermission(Permission.CONFIGURE);
```
yaroslavafenkin, line 155: This will be incorrect for some configurations. Please use `@AncestorInPath` here to obtain the context, and then check for `Item.CONFIGURE` permission on that context. See https://www.jenkins.io/doc/developer/security/form-validation/#checking-permissions for example.

yaroslavafenkin (changes requested):

src/main/java/com/mathworks/ci/MatlabBuilder.java

src/main/java/com/mathworks/ci/MatlabBuilder.java

yaroslavafenkin (approved):
Looks fine from security PoV.

PR #276, opened 2023-12-12
Update startup options to remove extraneous spaces

sameagen-MW:
Same change as the TeamCity plugin.


sameagen-MW:
I don't think this change is anywhere large enough to warrant its own release, but it seems like it would be good to get it in as a quality of life change whenever we do end up next releasing.

Vahila:
Hi Sam, JFYI, Jenkins build is passing when the inputs are specified with whitespaces. I tried the same command as in the TeamCity plugin and the Jenkins build is running successfully.

sameagen-MW:
Awesome, thanks for letting me know! `ArgumentListBuilder` must already take care of removing empty arguments. I'll go ahead and close the pull request then as it's unnecessary.

PR #277, opened 2023-12-15
Jenkins SD fix

sameagen-MW:
Verified interactively for all pipeline and UI steps.


davidbuzinski (approved):

nbhoski:
It still uses `cd` would not thet be an issue may be i dont  understand how cd (getenv(path)) is different from cd(path).

sameagen-MW:
Hey Nikhil, this is to fix a pretty specific issue where if the user passes in the `-sd` option it will currently be effectively ignored because we always change back to the workspace directory rather than the directory that MATLAB was started in.

I'll change the commit to move it over to `dev_main`.

nbhoski (approved):

PR #304, opened 2024-03-21
adding a new screenshot for build results and minor updates

mw-hrastega:


nbhoski (approved):
Looks good to me. I thought we decided to show a summary page screenshot instead?
I was just wondering if the table should also contain one Failure and skipped task just to cover all possible scenarios as FAILED status also contains the link to the task log in the console output.

mw-hrastega:
> Looks good to me. I thought we decided to show a summary page screenshot instead? I was just wondering if the table should also contain one Failure and skipped task just to cover all possible scenarios as FAILED status also contains the link to the task log in the console output.

Thanks for your comment, Nikhil. I followed up by removing a phrase that could be misleading. Regarding the screenshot, I guess it's better to show a desirable outcome (all tasks passing). I had to go with the table in the screenshot instead of the entry point view, because the entry point view includes the date and username and is not that doc-friendly. The table of task results should be able to communicate the feature as well.

davidbuzinski (approved):

Issue #326, opened 2024-06-11
Windows server doesn't work on mac or linux agent


yitingw3:
I have a Jenkins server running on a windows machine with a mac agent and a linux agent.
Got the error below on both mac and linux agent.

#################### Starting command output ####################
$ ********
Generating MATLAB script with content:
pwd

[mac_free] $ /Users/yitingw/bash/jenkins/workspace/mac_free\.matlab\tempDir.tmp16468002313306274218\run-matlab-command "setenv('MW_ORIG_WORKING_FOLDER', cd('/Users/yitingw/bash/jenkins/workspace/mac_free\.matlab\tempDir.tmp16468002313306274218'));script_WrTxnJhq" -sd matlab
Cannot run program "/Users/yitingw/bash/jenkins/workspace/mac_free\.matlab\tempDir.tmp16468002313306274218\run-matlab-command" (in directory "/Users/yitingw/bash/jenkins/workspace/mac_free"): error=13, Permission denied
Finished: FAILURE

PR #329, opened 2024-06-11
Fix 326

sameagen-MW:
Fix issue where the wrong file separator was used in cases where the Jenkins server was a Windows machine and the agents were Unix and vice-versa.


nbhoski (approved):
Shipt it !

nbhoski:
Closes #326

Issue #332, opened 2024-06-20
Files from '.matlab' folder are not deleted


Vahila:
Files from '.matlab' folder are not deleted. This affects the code coverage results when user doesn't specify the source folder.

sameagen-MW:
Confirmed to occur in `dev_main`

sameagen-MW:
It seems like we could do something strange like deleting the file from within itself once it's loaded into memory, but I think my preference would probably just be to add in the restriction that a folder(s) must be specified to collect code coverage.

sameagen-MW:
I don't know, I"m pretty conflicted here.

Vahila:
> Hey Vahila, sorry that I'm just now getting to look into this. Ended up getting wrapped up into a couple of other things for the last few days.
> 
> To make sure that I understand the issue correctly, the temporary script that's generated in the temp folder to run tests is being included within code coverage results if the user doesn't specify the source directory?
> 
> I don't think that's a bug, as we can't delete the script before generating results because the script is what's actually running the tests and collecting the coverage.
> 
> @nikhilbhoski and @mcafaro, what are your thoughts here? Are there any measures that we can take to get around this like a way to always exclude the `.matlab` folder when collecting coverage?

Hi Sam, yes, that is the issue here where the coverage metrics includes the files from '.matlab' folder when the user doesn't specify the source directory. 

This would be an unexpected behavior from user perspective as the source folder option is optional and also this include files that are generated by the plugin and are not related to the user. So I think we need to handle it somehow(deleting the files or excluding the folder)

sameagen-MW:
I'll keep thinking about it, I'm not sure that I'm totally convinced either way yet.

In the meantime, I do think the cleanup can be enhanced since at the moment it's tied to Java shutdown hooks which isn't ideal if the JVM instance is long-lived. I'll get that updated while I think about the code coverage issue.

Vahila:
This issue is resolved in 2.14.0 version

PR #334, opened 2024-06-26
Improve teardown

sameagen-MW:
Not a fix for issue [332](https://github.com/mathworks/jenkins-matlab-plugin/issues/332). I still need to think through what the consequences for having the temporary scripts self destruct would be. E.g. would we be resilient to page faults at high load, does this severely hinder debuggability, etc...

However, this should slightly improve performance as now the `MatlabCommandRunner` objects can be cleaned up before JVM shutdown. Especially for very long running Jenkins servers / agents this should be an improvement.

### DIFF
--- a/src/main/java/com/mathworks/ci/actions/RunMatlabBuildAction.java
+++ b/src/main/java/com/mathworks/ci/actions/RunMatlabBuildAction.java
@@ -81,6 +81,14 @@ public class RunMatlabBuildAction {
             throw(e);
         } finally {
             annotator.forceEol();
+
+            try {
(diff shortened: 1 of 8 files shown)


nbhoski:
> However, this should slightly improve performance as now the `M`

Page fault may be remote occurrence in this case as I would see this anyways cleaning the the acquired space. Ofr now I feel garbage collection through finally seems reasonable per build.

nbhoski (approved):
Shitp it !

davidbuzinski (approved):

PR #335, opened 2024-06-27
Switch to using a temporary directory on the agent external to the WS.

sameagen-MW:
This should address the issue with the code coverage, without needing to delete the generated scripts + binary until the end of the step when they are no longer needed.

If a the workspace is `//some/path/workspace`, now the files will be copied into `//some/path/workspace@tmp/matlab-uuid`. Currently they are copied into `//some/path/workspace/.matlab/tempDir-uuid`.

Removed the integration tests checking for a `.matlab` folder because checking that the temp directory is created properly and in the right place is already covered by the unit tests and is implementation detail rathern than part of the interface that we need to remain invariant.

### DIFF
--- a/src/main/java/com/mathworks/ci/BuildArtifactAction.java
+++ b/src/main/java/com/mathworks/ci/BuildArtifactAction.java
@@ -60,7 +60,7 @@ public class BuildArtifactAction implements Action {
 
     public List<BuildArtifactData> getBuildArtifact() throws ParseException, InterruptedException, IOException {
         List<BuildArtifactData> artifactData = new ArrayList<BuildArtifactData>();
-        FilePath fl = new FilePath(new File(build.getRootDir().getAbsolutePath() + "/" + BUILD_ARTIFACT_FILE));
+        FilePath fl = new FilePath(new File(build.getRootDir().getAbsolutePath(), BUILD_ARTIFACT_FILE));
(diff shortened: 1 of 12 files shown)


nbhoski:
What is the code coverage issue here ? Do we have any issue created . sorry I am not getting the context why we might need this

davidbuzinski:
Will want to double check that this still works if used inside a container on a Docker agent.

sameagen-MW:
Definitely still works on a Jenkinsfile Docker agent. It seems like all Docker integrations into Jenkins end up just mounting the workspace (not project) as a volume so there shouldn't be any issues from a theoretical sense. 

There's still some aspects that I need to check play nicely though like generating the build summary and code coverage / results reports. I'll switch the pr over to a draft since we're going to be out for the next almost-week.

davidbuzinski:
> Definitely still works on a Jenkinsfile Docker agent. It seems like all Docker integrations into Jenkins end up just mounting the workspace (not project) as a volume so there shouldn't be any issues from a theoretical sense.
> 
> There's still some aspects that I need to check play nicely though like generating the build summary and code coverage / results reports. I'll switch the pr over to a draft since we're going to be out for the next almost-week.

Sounds good, thanks for checking that out!

davidbuzinski (approved):

nbhoski (approved):
Ship it !

PR #348, opened 2024-07-23
Issue 345

### LINKED ISSUE #345: 'MATLAB Build Results' links from 2.1.3.1 throws a warning in the system log
When trying to access the Build results which are generated in 2.13.1, throws a warning.

Repro steps:
1. Install 2.13.1 version and generate a Build result table
2. Update the plugin to new changes from dev_main

Anomalies with accessing the 'MATLAB Build Results':

When clicked on 'MATLAB Build Results' :
image
It shows how many tasks are run but not the tasks information:
image

Clicking on 'MATLAB Build Results'  also throws a warning in Jenkins System log:
image

When clicked on the 'MATLAB Build Results' from build Status page, no table is shown
image

What is the expected behavior with build results form previous version?

nbhoski:
This PR closes #345 
* Maintains the backward compatibility for reading build artifact

### DIFF
--- a/src/main/java/com/mathworks/ci/BuildArtifactAction.java
+++ b/src/main/java/com/mathworks/ci/BuildArtifactAction.java
@@ -43,7 +43,7 @@ public class BuildArtifactAction implements Action {
     }
 
     public String getActionID(){
-        return this.actionID;
+        return (this.actionID == null) ? "" : this.actionID;
     }
 
     @CheckForNull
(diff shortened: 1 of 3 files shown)


src/main/java/com/mathworks/ci/BuildArtifactAction.java
```diff
@@ -61,14 +61,19 @@ public String getDisplayName() {
     @CheckForNull
     @Override
     public String getUrlName() {
-        return "buildresults" + this.actionID ;
+        return (this.actionID == null) ? "buildresults" : "buildresults" + this.actionID ;
```
mw-kapilg, line 64:
```suggestion
        return "buildresults" + (this.actionID == null ? "" : this.actionID);
```

src/main/java/com/mathworks/ci/BuildArtifactAction.java
mw-kapilg, line 1: Should we just initialize the class variable `actionsID` as empty string like `private String actionID = "";` (line 27)? That way we'd able to avoid all the null checking here

mw-kapilg (approved):

src/main/java/com/mathworks/ci/BuildArtifactAction.java
nbhoski: hmm, We would still need to check for ""

src/main/java/com/mathworks/ci/BuildArtifactAction.java
nbhoski: why to send buildresults+"" ?

nbhoski:
Thanks @mw-kapilg

src/main/java/com/mathworks/ci/BuildArtifactAction.java
mw-kapilg: We could use the `actionID` variable directly right?

src/main/java/com/mathworks/ci/BuildArtifactAction.java
mw-kapilg: The suggested change is to avoid using the string literal `"buildresults"` twice

PR #353, opened 2024-08-06
Added check for command textbox

nbhoski:
Fixed geck # 2221893 

* `Command` text box under `Run MATLAB Command` now checks if the the command box is empty 
* Empty command box shows an error message 

image

* Added Integ tests

### DIFF
--- a/src/main/java/com/mathworks/ci/freestyle/RunMatlabCommandBuilder.java
+++ b/src/main/java/com/mathworks/ci/freestyle/RunMatlabCommandBuilder.java
@@ -7,10 +7,13 @@ package com.mathworks.ci.freestyle;
  * 
  */
 
+import hudson.util.FormValidation;
 import java.io.IOException;
 import javax.annotation.Nonnull;
+import jenkins.model.Jenkins;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.DataBoundSetter;
+import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.StaplerRequest;
 import hudson.EnvVars;
 import hudson.Extension;
@@ -33,6 +36,7 @@ import com.mathworks.ci.parameters.RunActionParameters;
 import com.mathworks.ci.actions.MatlabActionFactory;
 import com.mathworks.ci.actions.RunMatlabCommandAction;
 import com.mathworks.ci.freestyle.options.StartupOptions;
+import org.kohsuke.stapler.verb.POST;
 
 public class RunMatlabCommandBuilder extends Builder implements SimpleBuildStep {
     // Deprecated
@@ -109,6 +113,14 @@ public class RunMatlabCommandBuilder extends Builder implements SimpleBuildStep 
         public boolean isApplicable(
                 @SuppressWarnings("rawtypes") Class<? extends AbstractProject> jobtype) {
             return true;
+        }
+        @POST
+        public FormValidation doCheckMatlabCommand(@QueryParameter String matlabCommand) {
+            Jenkins.get().checkPermission(Jenkins.ADMINISTER);
+            if (matlabCommand.isEmpty()) {
+                return FormValidation.error(Message.getValue("matlab.empty.command.error"));
+            }
+            return FormValidation.ok();
         }
     }
 
--- a/src/main/java/com/mathworks/ci/freestyle/RunMatlabCommandBuilder.java
+++ b/src/main/java/com/mathworks/ci/freestyle/RunMatlabCommandBuilder.java
@@ -116,9 +116,9 @@ public class RunMatlabCommandBuilder extends Builder implements SimpleBuildStep 
         }
 
         @POST
-        public FormValidation doCheckMatlabCommand(@QueryParameter String matlabCommand) {
+        public FormValidation doCheckMatlabCommand(@QueryParameter String value) {
             Jenkins.get().checkPermission(Jenkins.ADMINISTER);
-            if (matlabCommand.isEmpty()) {
+            if (value.isEmpty()) {
                 return FormValidation.error(Message.getValue("matlab.empty.command.error"));
             }
             return FormValidation.ok();
--- a/src/main/java/com/mathworks/ci/freestyle/RunMatlabCommandBuilder.java
+++ b/src/main/java/com/mathworks/ci/freestyle/RunMatlabCommandBuilder.java
@@ -114,6 +114,7 @@ public class RunMatlabCommandBuilder extends Builder implements SimpleBuildStep 
                 @SuppressWarnings("rawtypes") Class<? extends AbstractProject> jobtype) {
             return true;
         }
+
         @POST
         public FormValidation doCheckMatlabCommand(@QueryParameter String matlabCommand) {
             Jenkins.get().checkPermission(Jenkins.ADMINISTER);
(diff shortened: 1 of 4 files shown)


src/main/java/com/mathworks/ci/freestyle/RunMatlabCommandBuilder.java
```diff
@@ -110,6 +114,14 @@ public boolean isApplicable(
...
             return true;
         }
+        @POST
+        public FormValidation doCheckMatlabCommand(@QueryParameter String matlabCommand) {
+            Jenkins.get().checkPermission(Jenkins.ADMINISTER);
+            if (matlabCommand.isEmpty()) {
+                return FormValidation.error(Message.getValue("matlab.empty.command.error"));
+            }
+            return FormValidation.ok();
+        }
```
sameagen-MW, line 124:
```suggestion
        public FormValidation doCheckMatlabCommand(@QueryParameter String value) {
            Jenkins.get().checkPermission(Jenkins.ADMINISTER);
            if (value.isEmpty()) {
                return FormValidation.error(Message.getValue("matlab.empty.command.error"));
            }
            return FormValidation.ok();
        }

```

sameagen-MW: I believe the parameter name is significant here, and needs to be `value`. See [this](https://pitchfork.com/reviews/albums/britney-spears-blackout/) page for reference.

src/main/java/com/mathworks/ci/freestyle/RunMatlabCommandBuilder.java
```diff
@@ -110,6 +114,14 @@ public boolean isApplicable(
                 @SuppressWarnings("rawtypes") Class<? extends AbstractProject> jobtype) {
             return true;
         }
+        @POST
```
sameagen-MW, line 117: I don't think that this is necessary. Also, we should add a blank line above it for whitespace consistentcy.

src/main/resources/com/mathworks/ci/freestyle/RunMatlabCommandBuilder/config.jelly
```diff
@@ -2,7 +2,7 @@
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
     
 	  <f:entry title="Command" field="matlabCommand">
-	        <f:textbox/>
+	        <f:textbox checkMethod="post"/>
```
sameagen-MW, line 5: I don't think that this should be necessary. See [here](https://pitchfork.com/reviews/albums/britney-spears-blackout/).

src/test/java/integ/com/mathworks/ci/RunMatlabCommandBuilderTest.java
```diff
@@ -7,6 +7,9 @@
  * 
  */
 
+import com.gargoylesoftware.htmlunit.WebAssert;
+import com.gargoylesoftware.htmlunit.html.HtmlCheckBoxInput;
+import com.gargoylesoftware.htmlunit.html.HtmlPage;
```
sameagen-MW, line 12: I see that the pom file hasn't been updated. Is this already a Jenkins dependency?

src/main/java/com/mathworks/ci/freestyle/RunMatlabCommandBuilder.java
```diff
@@ -110,6 +114,14 @@ public boolean isApplicable(
                 @SuppressWarnings("rawtypes") Class<? extends AbstractProject> jobtype) {
             return true;
         }
+        @POST
+        public FormValidation doCheckMatlabCommand(@QueryParameter String matlabCommand) {
+            Jenkins.get().checkPermission(Jenkins.ADMINISTER);
```
sameagen-MW, line 119: Is there any reason why we're checking this? I would expect that Jenkins should already handle the permissions of who can edit a pipeline without us needing to do that.

sameagen-MW (changes requested):
This looks slick. The integration tests especially look great. There's just a couple of things that I'm not sure are totally necessary or at least that I don't fully understand at the moment.

src/main/java/com/mathworks/ci/freestyle/RunMatlabCommandBuilder.java
nbhoski: Do you mean [this](https://www.jenkins.io/doc/developer/security/form-validation/) link

nbhoski: I had named it to be in synch with our `UseBuildWrapper` class validations https://github.com/mathworks/jenkins-matlab-plugin/blob/a66bd3a3262d42937074bb783871456986c6d2b5/src/main/java/com/mathworks/ci/UseMatlabVersionBuildWrapper.java#L145

src/main/java/com/mathworks/ci/freestyle/RunMatlabCommandBuilder.java
nbhoski: This Recommended approach for avoiding CSRF attacks please refer [this](https://www.jenkins.io/doc/developer/security/form-validation/#protecting-from-csrf)  also we had received the Jenkins issue from Jenkins maintainers earlier to use this approach from security perspective

src/main/resources/com/mathworks/ci/freestyle/RunMatlabCommandBuilder/config.jelly
nbhoski: same as above the link is broken i have referred [this](https://www.jenkins.io/doc/developer/security/form-validation/#protecting-from-csrf)

src/main/java/com/mathworks/ci/freestyle/RunMatlabCommandBuilder.java
nbhoski: Same as above the read permission is implied hence such static methods are susceptible for security issues.

src/test/java/integ/com/mathworks/ci/RunMatlabCommandBuilderTest.java
nbhoski: Yes , We already have HTML unit tests and dependencies are part of it.

src/main/java/com/mathworks/ci/freestyle/RunMatlabCommandBuilder.java
sameagen-MW: Haha, sorry about that link. I think I was talking about music with a friend right about the same time I was writing the review. [This](https://www.jenkins.io/doc/developer/forms/form-validation/) was the link that I meant to send. 

 I think that you're right that there's no downside to taking the more secure option here, so you can just ignore this comment chain.

sameagen-MW (changes requested):
Looks good, I think the only thing I see is that we probably need to either change the validation function parameter name to `value` or use `@QueryParameter("value")`, as outlined [here](https://www.jenkins.io/doc/developer/forms/form-validation/#form-validation).

src/main/java/com/mathworks/ci/freestyle/RunMatlabCommandBuilder.java
```diff
@@ -110,6 +114,15 @@ public boolean isApplicable(
                 @SuppressWarnings("rawtypes") Class<? extends AbstractProject> jobtype) {
             return true;
         }
+
+        @POST
+        public FormValidation doCheckMatlabCommand(@QueryParameter String matlabCommand) {
```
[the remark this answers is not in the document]
nbhoski: Done

sameagen-MW (approved):

Issue #285, opened 2024-01-31
The hyperlink of the task doesn't point to the logs correctly when the tasks are skipped


Vahila:
When the tasks in the buildfile are skipped, the hyperlink of the task doesn't point to the build task logs. It just opens the starting of that Jenkins build log.

Reproduction steps:
Buildfile.m
function plan = buildfile
% Create a plan from the task functions
plan = buildplan(localfunctions);

% Specify the inputs and outputs of the "pcode" task
plan("pcode").Inputs = "source/**/*.m";
plan("pcode").Outputs = plan("pcode").Inputs.replace(".m",".p");

% Specify the inputs and outputs of the "archive" task
plan("archive").Inputs = plan("pcode").Outputs;
plan("archive").Outputs = "source.zip";
end

function pcodeTask(context)
% Create P-code files
filePaths = context.Task.Inputs.paths;
pcode(filePaths{:},"-inplace")
end

function archiveTask(context)
% Create ZIP file
task = context.Task;
zip(task.Outputs.paths,task.Inputs.paths)
end

- Add the build file in your workspace. Create a folder 'source' in the workspace.
- Run the 'archive' task twice.
- In the second run, both the tasks 'pcode' and 'archive' are skipped.
- When the links from the MATLAB Build Results page(Hyperlinks to the task names) are clicked, it points to the start of Jenkins build log page not to the start of the buildtool task logs.

Issue #337, opened 2024-06-30
Skipped build tasks do not have a working link


acampbel:
There is no link for skipped build tasks using the Run MATLAB build step.

nbhoski:
Yes, this was known issue with skip it does not enter into our plugin code hence we currently wont be able to add our sentinel for skip tasks also second rationale was, it would not bemuch helpful to show just skip as there would not be any stack trace or reason to skip will be displayed in the log.

mcafaro:
There is a `skipTask` method on `BuildRunnerPlugin` that is invoked whenever a task will be skipped.

PR #279, opened 2024-01-18
2.11.2 snapshot qualification

nbhoski:
* Adds support for MATLAB Build visulization both for Freestyle and pipeline 
*

PR #354, opened 2024-08-09
Add build results support in Run Command step and skipped task logs link in table

mw-kapilg:
Fixes https://github.com/mathworks/jenkins-matlab-plugin/issues/285, https://github.com/mathworks/jenkins-matlab-plugin/issues/337 and https://github.com/mathworks/jenkins-matlab-plugin/issues/349 by:

- Adding sentinel for skipped tasks in logs and linking table links to them
- Adding build results table generation for Run MATLAB command step

### DIFF
--- a/src/main/java/com/mathworks/ci/actions/RunMatlabBuildAction.java
+++ b/src/main/java/com/mathworks/ci/actions/RunMatlabBuildAction.java
@@ -11,6 +11,7 @@ import java.io.IOException;
 import hudson.FilePath;
 import hudson.model.Run;
 
+import com.mathworks.ci.MatlabBuilderConstants;
 import com.mathworks.ci.BuildArtifactAction;
 import com.mathworks.ci.BuildConsoleAnnotator;
 import com.mathworks.ci.MatlabExecutionException;
@@ -22,13 +23,6 @@ public class RunMatlabBuildAction {
     private BuildActionParameters params; 
     private MatlabCommandRunner runner;
     private BuildConsoleAnnotator annotator;
-
-    private static String DEFAULT_PLUGIN = 
-        "+ciplugins/+jenkins/getDefaultPlugins.m";
-    private static String BUILD_REPORT_PLUGIN = 
-        "+ciplugins/+jenkins/BuildReportPlugin.m";
-    private static String TASK_RUN_PROGRESS_PLUGIN = 
-        "+ciplugins/+jenkins/TaskRunProgressPlugin.m";
     private String actionID;
 
     public String getActionID(){
@@ -52,9 +46,9 @@ public class RunMatlabBuildAction {
 
     public void run() throws IOException, InterruptedException, MatlabExecutionException {
         // Copy plugins and override default plugins function
-        runner.copyFileToTempFolder(DEFAULT_PLUGIN, DEFAULT_PLUGIN);
-        runner.copyFileToTempFolder(BUILD_REPORT_PLUGIN, BUILD_REPORT_PLUGIN);
-        runner.copyFileToTempFolder(TASK_RUN_PROGRESS_PLUGIN, TASK_RUN_PROGRESS_PLUGIN);
+        runner.copyFileToTempFolder(MatlabBuilderConstants.DEFAULT_PLUGIN, MatlabBuilderConstants.DEFAULT_PLUGIN);
+        runner.copyFileToTempFolder(MatlabBuilderConstants.BUILD_REPORT_PLUGIN, MatlabBuilderConstants.BUILD_REPORT_PLUGIN);
+        runner.copyFileToTempFolder(MatlabBuilderConstants.TASK_RUN_PROGRESS_PLUGIN, MatlabBuilderConstants.TASK_RUN_PROGRESS_PLUGIN);
 
         
         // Set environment variable
@@ -119,6 +113,5 @@ public class RunMatlabBuildAction {
                 }
             }
         }
-
     }
 }
--- a/src/main/java/com/mathworks/ci/actions/RunMatlabCommandAction.java
+++ b/src/main/java/com/mathworks/ci/actions/RunMatlabCommandAction.java
@@ -1,5 +1,7 @@
 package com.mathworks.ci.actions;
 
+import java.io.File;
+
 /**
  * Copyright 2024, The MathWorks Inc.
  *
@@ -7,37 +9,100 @@ package com.mathworks.ci.actions;
 
 import java.io.IOException;
 
+import org.apache.commons.lang.RandomStringUtils;
+
+import com.mathworks.ci.BuildArtifactAction;
+import com.mathworks.ci.BuildConsoleAnnotator;
+import com.mathworks.ci.MatlabBuilderConstants;
 import com.mathworks.ci.MatlabExecutionException;
 import com.mathworks.ci.parameters.RunActionParameters;
 import com.mathworks.ci.utilities.MatlabCommandRunner;
 
+import hudson.FilePath;
+import hudson.model.Run;
+
 public class RunMatlabCommandAction {
     private RunActionParameters params; 
     private MatlabCommandRunner runner;
+    private BuildConsoleAnnotator annotator;
+    private String actionID;
 
-    public RunMatlabCommandAction(MatlabCommandRunner runner, RunActionParameters params) {
+    public String getActionID(){
+        return this.actionID;
+    }
+
+    public RunMatlabCommandAction(MatlabCommandRunner runner, BuildConsoleAnnotator annotator, RunActionParameters params) {
         this.runner = runner;
+        this.actionID = RandomStringUtils.randomAlphanumeric(8);
+        this.annotator = annotator;
         this.params = params;
     }
 
     public RunMatlabCommandAction(RunActionParameters params) throws IOException, InterruptedException {
-        this(new MatlabCommandRunner(params), params);
+        this(new MatlabCommandRunner(params), 
+                new BuildConsoleAnnotator(
+                    params.getTaskListener().getLogger(), 
+                    params.getBuild().getCharset()),
+                params);
     }
 
     public void run() throws IOException, InterruptedException, MatlabExecutionException {
+        // Copy plugins and override default plugins function
+        runner.copyFileToTempFolder(MatlabBuilderConstants.DEFAULT_PLUGIN, MatlabBuilderConstants.DEFAULT_PLUGIN);
+        runner.copyFileToTempFolder(MatlabBuilderConstants.BUILD_REPORT_PLUGIN, MatlabBuilderConstants.BUILD_REPORT_PLUGIN);
+        runner.copyFileToTempFolder(MatlabBuilderConstants.TASK_RUN_PROGRESS_PLUGIN, MatlabBuilderConstants.TASK_RUN_PROGRESS_PLUGIN);
+
+        
+        // Set environment variable
+        runner.addEnvironmentVariable(
+                "MW_MATLAB_BUILDTOOL_DEFAULT_PLUGINS_FCN_OVERRIDE",
+                "ciplugins.jenkins.getDefaultPlugins");
+        runner.addEnvironmentVariable("MW_BUILD_PLUGIN_ACTION_ID",this.getActionID());
+        runner.addEnvironmentVariable(
+                "MW_MATLAB_TEMP_FOLDER",
+                runner.getTempFolder().toString());
+
+        // Redirect output to the build annotator
+        runner.redirectStdOut(annotator);
+
+        // Prepare MATLAB command
+        String command = "addpath('" 
+            + runner.getTempFolder().getRemote()
+            + "');" + this.params.getCommand();
+
         try {
-            runner.runMatlabCommand(this.params.getCommand());
+            runner.runMatlabCommand(command);
         } catch (Exception e) {
             this.params.getTaskListener().getLogger()
                 .println(e.getMessage());
             throw(e);
         } finally {
+            annotator.forceEol();
+
             try {
-                this.runner.removeTempFolder();
+                // Handle build result
+                Run<?,?> build = this.params.getBuild();
+                FilePath jsonFile = new FilePath(runner.getTempFolder(), "buildArtifact.json");
+                if (jsonFile.exists()) {
+                    FilePath rootLocation = new FilePath(
+                            new File(
+                                build.getRootDir().getAbsolutePath(),
+                                "buildArtifact" + this.getActionID() + ".json")
+                            );
+                    jsonFile.copyTo(rootLocation);
+                    jsonFile.delete();
+                    build.addAction(new BuildArtifactAction(build, this.getActionID()));
+                }
             } catch (Exception e) {
                 // Don't want to override more important error
                 // thrown in catch block
                 System.err.println(e.toString());
+            } finally {
+                try {
+                    this.runner.removeTempFolder();
+                } catch (Exception e) {
+                    System.err.println(e.toString());
+                }
             }
         }
     }
--- a/src/main/resources/com/mathworks/ci/BuildArtifactAction/summary.jelly
+++ b/src/main/resources/com/mathworks/ci/BuildArtifactAction/summary.jelly
@@ -12,12 +12,15 @@
         <font color="#EF2929"><h5>Unable to generate a build artifact. </h5></font>
       </j:if>
     </span>
-    <p><b>Tasks run: <font color="Blue">${it.totalCount}</font></b></p>
-    <br>
-    <b>Failed: <font color="red">${it.failCount}</font></b>
-    </br>
-    <br>
-    <b>Skipped: <font color="orange">${it.skipCount}</font></b>
-    </br>
+    <p>
+      <b>Tasks run: <font color="Blue">${it.totalCount}</font></b>
+      <br></br>
+      <br>
+      <b>Failed: <font color="red">${it.failCount}</font></b>
+      </br>
+      <br>
+      <b>Skipped: <font color="orange">${it.skipCount}</font></b>
+      </br>
+    </p>
   </t:summary>
 </j:jelly>
--- a/src/test/java/unit/com/mathworks/ci/actions/RunMatlabBuildActionTest.java
+++ b/src/test/java/unit/com/mathworks/ci/actions/RunMatlabBuildActionTest.java
@@ -18,6 +18,9 @@ import static org.junit.Assert.*;
 import org.mockito.Mock;
 import org.mockito.InOrder;
 import org.mockito.junit.MockitoJUnitRunner;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.*;
 
 import hudson.FilePath;
@@ -26,7 +29,9 @@ import hudson.model.TaskListener;
 
 import com.mathworks.ci.BuildArtifactAction;
 import com.mathworks.ci.BuildConsoleAnnotator;
+import com.mathworks.ci.MatlabBuilderConstants;
 import com.mathworks.ci.MatlabExecutionException;
+import com.mathworks.ci.actions.RunMatlabBuildAction;
 import com.mathworks.ci.utilities.MatlabCommandRunner;
 import com.mathworks.ci.parameters.BuildActionParameters;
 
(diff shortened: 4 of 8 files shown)


nbhoski:
@mw-kapilg are you working on the pipeline failures? I shall add my review after you these fixed

src/main/resources/com/mathworks/ci/BuildArtifactAction/summary.jelly
```diff
@@ -12,12 +12,15 @@
...
       </j:if>
     </span>
-    <p><b>Tasks run: <font color="Blue">${it.totalCount}</font></b></p>
-    <br>
-    <b>Failed: <font color="red">${it.failCount}</font></b>
-    </br>
-    <br>
-    <b>Skipped: <font color="orange">${it.skipCount}</font></b>
-    </br>
+    <p>
```
nbhoski, line 15: @mw-hrastega could verify this agian i remember some changes here suggested by you here 
https://github.com/mathworks/jenkins-matlab-plugin/pull/279/commits/2203ba66c9a06b67936d182c6dddf6ed810b181f

src/test/java/unit/com/mathworks/ci/actions/RunMatlabBuildActionTest.java
```diff
@@ -65,21 +70,21 @@ public void init() {
     public void shouldCopyPluginsToTempDirectory() throws IOException, InterruptedException, MatlabExecutionException {
         action.run();
 
-        String DEFAULT_PLUGIN = 
-            "+ciplugins/+jenkins/getDefaultPlugins.m";
-        String BUILD_REPORT_PLUGIN = 
-            "+ciplugins/+jenkins/BuildReportPlugin.m";
-        String TASK_RUN_PROGRESS_PLUGIN = 
-            "+ciplugins/+jenkins/TaskRunProgressPlugin.m";
+        // String DEFAULT_PLUGIN =
```
nbhoski, line 73: We can remove these comments

src/test/java/unit/com/mathworks/ci/actions/RunMatlabCommandActionTest.java
```diff
@@ -38,17 +55,55 @@ public class RunMatlabCommandActionTest {
...
     }
 
+    @Test
+    public void shouldCopyPluginsToTempDirectory() throws IOException, InterruptedException, MatlabExecutionException {
+        action.run();
+
+        InOrder inOrder = inOrder(runner);
+
+        inOrder.verify(runner)
+            .copyFileToTempFolder(MatlabBuilderConstants.DEFAULT_PLUGIN, MatlabBuilderConstants.DEFAULT_PLUGIN);
```
nbhoski, line 77: Same as above combining these might help us re writing it again.

nbhoski (changes requested):

src/main/java/com/mathworks/ci/actions/RunMatlabCommandAction.java
```diff
@@ -1,43 +1,108 @@
 package com.mathworks.ci.actions;
 
+import java.io.File;
```
sameagen-MW, line 3: This should probably go below the copyright with the other import statements for consistency.

src/main/resources/com/mathworks/ci/BuildArtifactAction/summary.jelly
mw-kapilg: Checked with Houman offline and he agrees with the change

src/main/java/com/mathworks/ci/actions/RunMatlabCommandAction.java
```diff
@@ -1,43 +1,108 @@
...
+                    params.getTaskListener().getLogger(), 
+                    params.getBuild().getCharset()),
+                params);
     }
 
     public void run() throws IOException, InterruptedException, MatlabExecutionException {
+        // Copy plugins and override default plugins function
+        runner.copyFileToTempFolder(MatlabBuilderConstants.DEFAULT_PLUGIN, MatlabBuilderConstants.DEFAULT_PLUGIN);
+        runner.copyFileToTempFolder(MatlabBuilderConstants.BUILD_REPORT_PLUGIN, MatlabBuilderConstants.BUILD_REPORT_PLUGIN);
+        runner.copyFileToTempFolder(MatlabBuilderConstants.TASK_RUN_PROGRESS_PLUGIN, MatlabBuilderConstants.TASK_RUN_PROGRESS_PLUGIN);
```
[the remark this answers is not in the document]
mw-kapilg: Added a new parent class `RunMatlabAction` which is inherited by other action classes. This parent class includes the common functionalities

src/test/java/unit/com/mathworks/ci/actions/RunMatlabCommandActionTest.java
mw-kapilg: Moved the test to `RunMatlabActionTest.java`

src/main/java/com/mathworks/ci/actions/RunMatlabCommandAction.java
mw-kapilg: Thanks! I've updated for all other files as well

src/test/java/unit/com/mathworks/ci/actions/RunMatlabBuildActionTest.java
```diff
@@ -130,40 +107,4 @@ public void shouldPrintAndRethrowMessage() throws IOException, InterruptedExcept
...
-        // Should have copied file to root dir
-        assertTrue(new File(dest, "buildArtifact"+ action.getActionID() + ".json").exists());
-    }
-
-    @Test
-    public void shouldRemoveTempFolder() throws IOException, InterruptedException, MatlabExecutionException {
-        action.run();
-
-        verify(runner).removeTempFolder();
-    }
```
sameagen-MW, line 168: I don't think that this should be removed.

src/main/java/com/mathworks/ci/MatlabBuild.java
```diff
@@ -1,4 +1,5 @@
 package com.mathworks.ci;
+
```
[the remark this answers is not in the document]
sameagen-MW: Looks like I probably just missed marking it for delete in git.

src/main/java/com/mathworks/ci/FormValidationUtil.java
```diff
@@ -6,6 +6,7 @@
  * This is Utility class which provides commonly used methods for form validations across builders
  * 
  */
+
```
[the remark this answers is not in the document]
mw-kapilg: I see usage [here](https://github.com/mathworks/jenkins-matlab-plugin/blob/master/src/main/java/com/mathworks/ci/UseMatlabVersionBuildWrapper.java#L156), keeping it. Thanks for having a look!

src/main/java/com/mathworks/ci/MatlabBuild.java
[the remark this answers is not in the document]
mw-kapilg: Sure, I can remove it then

src/main/java/com/mathworks/ci/MatlabExecutionException.java
```diff
@@ -1,5 +1,10 @@
 package com.mathworks.ci;
 
+/**
+ * Copyright 2021 The MathWorks, Inc.
```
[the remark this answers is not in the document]
mw-kapilg: I had a look at previous releases and found it existed since 2021 ([commit history](https://github.com/mathworks/jenkins-matlab-plugin/commits/matlab-2.13.0/src/main/java/com/mathworks/ci/MatlabExecutionException.java) of file before refactoring)

src/test/java/unit/com/mathworks/ci/actions/RunMatlabBuildActionTest.java
mw-kapilg: Good catch, I unintentionally removed this test. Moving it to `RunMatlabActionTest.java`

src/main/java/com/mathworks/ci/MatlabBuild.java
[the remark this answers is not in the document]
sameagen-MW: Thanks for doing that!

src/main/java/com/mathworks/ci/actions/RunMatlabAction.java
```diff
@@ -0,0 +1,81 @@
...
+import com.mathworks.ci.MatlabBuilderConstants;
+import com.mathworks.ci.utilities.MatlabCommandRunner;
+import hudson.FilePath;
+import hudson.model.Run;
+import org.apache.commons.lang.RandomStringUtils;
+
+import java.io.File;
+import java.io.IOException;
+
+public class RunMatlabAction {
```
sameagen-MW, line 19: If this is going to be the parent class for the other three actions would you be open to naming it `MatlabAction` instead of `RunMatlabAction`? I think it's just a little bit cleaner.

src/main/java/com/mathworks/ci/actions/RunMatlabAction.java
```diff
@@ -0,0 +1,81 @@
...
+        runner.addEnvironmentVariable("MW_BUILD_PLUGIN_ACTION_ID",this.getActionID());
+        runner.addEnvironmentVariable(
+                "MW_MATLAB_TEMP_FOLDER",
+                runner.getTempFolder().toString());
+    }
+
+    public void moveArtifactToBuildRoot(Run<?,?> build, String artifactBaseName) {
+        try {
+            FilePath jsonFile = new FilePath(runner.getTempFolder(), artifactBaseName + ".json");
+            if (jsonFile.exists()) {
```
sameagen-MW, line 59: Should we add a check here to make sure that `this.getActionId()` doesn't return null?

src/main/java/com/mathworks/ci/actions/RunMatlabAction.java
```diff
@@ -0,0 +1,81 @@
...
+        runner.addEnvironmentVariable(
+                "MW_MATLAB_BUILDTOOL_DEFAULT_PLUGINS_FCN_OVERRIDE",
+                "ciplugins.jenkins.getDefaultPlugins");
+        runner.addEnvironmentVariable("MW_BUILD_PLUGIN_ACTION_ID",this.getActionID());
+        runner.addEnvironmentVariable(
+                "MW_MATLAB_TEMP_FOLDER",
+                runner.getTempFolder().toString());
+    }
+
+    public void moveArtifactToBuildRoot(Run<?,?> build, String artifactBaseName) {
```
sameagen-MW, line 56: I think that we should maybe try and figure out a name that also includes the fact that we're cleaning up the temp directory, because I think that's potentially even more important than moving the build artifact. I think it would help readability if it was clear to devs that were looking at the other files that this is a clean up function.

What do you think about `tearDownActions` or something in that vein? Then we could also use it in `RunMatlabTestsAction.java`

sameagen-MW, line 56: Also I think that we should remove `artifactBaseName` as a parameter, since we're just passing in that constant at every call site.

src/test/java/unit/com/mathworks/ci/actions/RunMatlabTestsActionTest.java
```diff
@@ -170,11 +170,4 @@ public void printsAndRethrowsMessage() throws IOException, InterruptedException,
             assertEquals(12, e.getExitCode());
         };
     }
-
-    @Test
-    public void shouldRemoveTempFolder() throws IOException, InterruptedException, MatlabExecutionException {
-        action.run();
-
-        verify(runner).removeTempFolder();
-    }
```
sameagen-MW, line 179: This test was removed, but I don't think that `RunMatlabTestsAction.java` was updated to use the new teardown method.

sameagen-MW (changes requested):
Thanks so much for working on this, I think that we're almost there. This ended up being a big maintenance update, so really appreciate you going through and keeping everything tidy.

src/main/java/com/mathworks/ci/actions/RunMatlabAction.java
mw-kapilg: I guess we could simply `MatlabAction` as well

src/main/java/com/mathworks/ci/actions/RunMatlabAction.java
mw-kapilg: Sure, wouldn't harm

src/main/java/com/mathworks/ci/actions/RunMatlabAction.java
mw-kapilg: I like the idea of having a `teardownActions()` method. I'll figure out some of the Exception related things and update the implementation

mw-kapilg: I was hoping for it to be a more generic method so that later when we want to move the `testResults` artifact, we could use the same method called with respective file name. Let me try a few things out otherwise I can keep it simple for now and defer conversion to generic method till I open the test results visualization PR

src/test/java/unit/com/mathworks/ci/actions/RunMatlabTestsActionTest.java
mw-kapilg: Oh yeah, I think we can start using the `teardownActions()` method in `RunMatlabTestsAction` as well. Will update it, thanks!

src/main/java/com/mathworks/ci/actions/RunMatlabAction.java
mw-kapilg: Added in the getter, similar to `BuildArtifactAction.getActionID()`

src/main/java/com/mathworks/ci/actions/MatlabAction.java
```diff
@@ -0,0 +1,83 @@
...
+                file.copyTo(rootLocation);
+                file.delete();
+                build.addAction(new BuildArtifactAction(build, this.getActionID()));
+            }
+        } catch (Exception e) {
+            // Don't want to override more important error
+            // thrown in catch block
+            System.err.println(e.toString());
+        }
+    }
```
sameagen-MW, line 82: Hmm, I'm a little bit nervous here about someone trying to call this method after using the constructor which does not provide an annotation. Could we check for the annotator being null and then throw an explicit error in that case rather than getting the null exception when it tries to concatenate `null` on line 66?

src/main/java/com/mathworks/ci/actions/RunMatlabBuildAction.java
```diff
@@ -93,32 +62,17 @@ public void run() throws IOException, InterruptedException, MatlabExecutionExcep
...
 
+    @Override
+    public void teardownAction() {
+        Run<?, ?> build = this.params.getBuild();
+        
+        // Handle build result
+        super.moveJsonArtifactToBuildRoot(build, MatlabBuilderConstants.BUILD_ARTIFACT);
+        
+        super.teardownAction();
     }
```
sameagen-MW, line 77: Definitely not a blocking piece of feedback here, just something I'm wondering about.

Do we want to override this here, or could we change the parent `teardownAction` to check if it has an annotator and then move the artifact if so? Overrides are generally not my preference, because it can lead to a pattern of changes having to be done in each child object which overrides a method.

Maybe something like this?

MatlabAction.java
```java
public void teardownAction() {
    if (this.annotator != null) {
        this.moveJsonArtifactToBuildRoot(...)
    }

    try {
        this.runner.removeTempFolder();
    } catch (Exception e) {
        System.err.println(e.toString());
    }
}
```

src/main/java/com/mathworks/ci/actions/RunMatlabCommandAction.java
```diff
@@ -7,38 +7,62 @@
...
+
+    @Override
+    public void teardownAction() {
+        Run<?, ?> build = this.params.getBuild();
+
+        // Handle build result
+        super.moveJsonArtifactToBuildRoot(build, MatlabBuilderConstants.BUILD_ARTIFACT);
+        
+        super.teardownAction();
+    }
```
sameagen-MW, line 67: This is kind of an example of the code duplication that I mentioned in my last comment that could get a little bit painful to maintain over time. Again definitely not a blocker.

sameagen-MW (approved):

nbhoski (changes requested):

.gitignore
```diff
@@ -7,4 +7,6 @@ src/main/resources/matlab-script-generator.zip
 src/main/resources/**/run-matlab-command*
 src/main/resources/license.txt
 
+.idea/
```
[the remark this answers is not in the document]
mw-kapilg: Yes, shouldn't be harmful to keep

src/main/java/com/mathworks/ci/MatrixPatternResolver.java
```diff
@@ -1,4 +1,5 @@
 package com.mathworks.ci;
+
```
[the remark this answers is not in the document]
mw-kapilg: So basically I have tried to follow this:

1st line -> `package ...`
2nd line -> blank
3rd line -> copyright

Both `MatrixPatterResolver.java` and `Message.java` follow this pattern. Hope that's fine, thanks!

src/main/java/com/mathworks/ci/actions/MatlabAction.java
mw-kapilg: Maybe I'm missing something but I'm unable to understand the concern here. Which part of line 66 are you expecting to equate to `null`? If I assume it's `runner.getTempFolder()`, all the Action class constructors require a runner and therefore `runner.getTempFolder` wouldn't give `null`.

P.S. For any other scenario as well if we need a `null` check, checking `null` for annotator would only solve for moving build artifact and **not** test results artifact (I mentioned in one of the previous comments about using this method for both build and test artifact in future)

src/main/java/com/mathworks/ci/actions/RunMatlabBuildAction.java
mw-kapilg: Even I wasn't completely convinced with this design but thought it made some sense here. Checking `null` for annotator could also be fine for now (atleast until we start using annotator somewhere in **Run MATLAB Tests** as well). If this feels better, I can update it.

@nbhoski do you have a preference here?

src/main/java/com/mathworks/ci/actions/RunMatlabTestsAction.java
nbhoski, line 1:
```suggestion
import java.io.IOException;
import java.util.List;
import java.util.ArrayList;
import hudson.FilePath;
import com.mathworks.ci.Utilities;
import com.mathworks.ci.MatlabBuilderConstants;
import com.mathworks.ci.MatlabExecutionException;
import com.mathworks.ci.parameters.TestActionParameters;
import com.mathworks.ci.utilities.MatlabCommandRunner;

```

src/main/java/com/mathworks/ci/actions/RunMatlabTestsAction.java
```diff
@@ -51,13 +50,7 @@ public void run() throws IOException, InterruptedException, MatlabExecutionExcep
...
             throw(e);
         } finally {
-            try {
-                this.runner.removeTempFolder();
-            } catch (Exception e) {
-                // Don't want to override more important error
-                // thrown in catch block
-                System.err.println(e.toString());
-            }
+            super.teardownAction();
```
nbhoski, line 53:
```suggestion
            teardownAction();
```

nbhoski (changes requested):

nbhoski:
I would suggest use of auto-formatting for inconsistent new lines for import statements and in the method description

src/main/java/com/mathworks/ci/actions/RunMatlabBuildAction.java
```diff
@@ -5,40 +5,21 @@
  *
  */
 
-import java.io.File;
 import java.io.IOException;
```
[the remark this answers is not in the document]
mw-kapilg: Discussed the formatting offline, so resolving this and following related comments

src/main/java/com/mathworks/ci/actions/RunMatlabTestsAction.java
mw-kapilg: I agree `super.teardown()` is not mandatory but slightly helps the reader so would prefer to keep it

Vahila (approved):
LGTM!

src/main/java/com/mathworks/ci/actions/MatlabAction.java
sameagen-MW: Sorry, you're totally right 66 is the wrong line number. I was thinking of the call to `this.getActionID()`, since that has the potential to return `null`. Might be worth handling just to make it harder to make mistakes in the codebase in the future like calling the wrong constructor.

If someone is ever just looking at the API for `MatlabAction.java` and doesn't pay enough attention to realize that `moveJsonArtifactToBuildRoot` is tied to one of two constructors it could be problematic.

mw-kapilg: We have a `null` check in the getter [here](https://github.com/mathworks/jenkins-matlab-plugin/pull/354/files/becf595f879c406a4450007329bdbc79b2c2f302#diff-b207e06a879c72043c0244c0f0b181868e031a2b275c265cfc339ada14f7e881R24-R26). I saw Nikhil had done similar for `BuildArtifactAction.java` earlier.

nbhoski (approved):
Ship it!

Vahila (approved):

PR #360, opened 2024-09-18
Add test results visualization

mw-kapilg:
This feature creates an in-platform view for MATLAB test results (whenever a default test runner or pre-made TestTask is used) if the user has a MATLAB Test license.

**Key highlights:**

- Each row of the table represents the test file.
- All tests for a single file are shown under that file.
- The files with same name and different parent directories are shown separately and you can hover over the file name to see its path.
- The table in table format allows easy management of test cases for a test file
- You see the logs in-place for a failed/incomplete. If there are more than failures within a test case, say multiple verification failures, each verification failure will have its own dropdown and log box.
- Each of the passed, failed, etc. buttons act as filters to respective type of tests
- The "Open in MATLAB Online" button helps to redirect users to MATLAB Online. If their Jenkins project is connected to a git repository, they are prompted to open the in MATLAB Online for easy debugging.
- The help button redirects users to our plugin's configuration guide.

Here's a few snapshots:
Screenshot 2024-09-18 113256
Screenshot 2024-09-18 113511


src/main/java/com/mathworks/ci/utilities/MatlabCommandRunner.java
```diff
@@ -34,20 +34,29 @@ public class MatlabCommandRunner {
...
+        if (tmpRoot == null) {
+            throw new IOException("Failed to create temporary directory");
+        }
         tmpRoot.mkdirs();
-
+    
         // Create temp folder
         this.tempFolder = tmpRoot.createTempDir("matlab", null);
+        if (this.tempFolder == null) {
+            throw new IOException("Failed to create MATLAB temporary directory");
```
mw-hrastega, line 58: Will this be user visible?

a temporary MATLAB directory

mw-kapilg: I believe so

mw-kapilg:
@mw-hrastega, for text review you can go through all the `.jelly` files.

src/main/java/com/mathworks/ci/utilities/MatlabCommandRunner.java
mw-hrastega: Ok, in this case, please follow these guidelines for all user-visible error messages: https://mathworks.sharepoint.com/:w:/s/devu/WritingErrorMessages/ERZSbiQk2fdFo4PNh30jK_4Bh_WDI6C-mSZIZYFCc_xumw?e=Gp0dOg

src/main/resources/com/mathworks/ci/TestResultsViewAction/index.jelly
```diff
@@ -0,0 +1,217 @@
...
+                </br><br></br>Failed</p>
+            </button>
+
+            <button id="Incomplete" onclick="showTests(id)" class="jenkins-button jenkins-!-margin-4 jenkins-!-padding-0" style="display: inline; width:8em;" tooltip="Incomplete tests">
+                <p>${it.incompleteCount}<br>
+                    <l:icon class="symbol-alert-circle-outline plugin-ionicons-api icon-lg jenkins-!-warning-color"/>
+                </br><br></br>Incomplete</p>
+            </button>
+
+            <button id="NotRun" onclick="showTests(id)" class="jenkins-button jenkins-!-margin-4 jenkins-!-padding-0" style="display: inline; width:8em;" tooltip="Not run tests">
```
mw-hrastega, line 36: Tests that have not run?

src/main/resources/com/mathworks/ci/TestResultsViewAction/index.jelly
```diff
@@ -0,0 +1,217 @@
...
+                    <l:icon class="symbol-remove-circle-outline plugin-ionicons-api icon-lg"/>
+                </br><br></br>Not Run</p>
+            </button>
+
+            <j:forEach var="testSession" items="${it.testResults}" varStatus="status">
+                <table class="jenkins-table sortable" id="testResultsTable">
+                    <thead>
+                        <tr>
+                            <th class="pane-header" style="width: auto; min-width: 20em;">
+                                Test Name
```
mw-hrastega, line 47: Test name should include the procedure name and information about parameterization. How about Test File instead?

src/main/resources/com/mathworks/ci/TestResultsViewAction/index.jelly
```diff
@@ -0,0 +1,217 @@
...
+                            </th>
+                            <th class="pane-header" style="width: auto; min-width: 10em; text-align: right;">
+                                Duration (seconds)
```
mw-hrastega, line 50: (s) instead of (seconds)?

mw-hrastega: Is it better to include + and - signs or drop-down arrows instead?

src/main/resources/com/mathworks/ci/TestResultsViewAction/index.jelly
```diff
@@ -0,0 +1,217 @@
...
+                                        <thead>
+                                            <tr>
+                                                <th class="pane-header" style="width: max-content;">
+                                                    Test Case
+                                                </th>
+                                                <th class="pane-header" style="width: 100%;">
+                                                    Diagnostics
+                                                </th>
+                                                <th class="pane-header" style="text-align: right; text-wrap:nowrap;">
+                                                    Duration (seconds) = ${testFile.duration}
```
mw-hrastega, line 95: = or :?

src/main/resources/com/mathworks/ci/TestResultsViewAction/index.jelly
```diff
@@ -0,0 +1,217 @@
...
+                                    </a>
+
+                                <div id="testCases-${testFile.name}-${testFile.id}" style="display: none;">
+                                    <br />
+                                    <br />
+                                    <table class="jenkins-table">
+                                        <thead>
+                                            <tr>
+                                                <th class="pane-header" style="width: max-content;">
+                                                    Test Case
```
mw-hrastega, line 89: Can't we use "Test" instead of "Test Case"?

src/main/resources/com/mathworks/ci/TestResultsViewAction/index.jelly
```diff
@@ -0,0 +1,217 @@
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:st="jelly:stapler" xmlns:c="/charts">
+    <l:layout title="MATLAB Test Results">
+        <l:side-panel>
+            <st:include page="sidepanel.jelly" it="${it.build}" optional="true" />
+        </l:side-panel>
+
+        <l:app-bar title="MATLAB Test Results">
+            <button class="jenkins-button" onclick="window.open('https://matlab.mathworks.com/open/github/v1?repo=${GIT_URL}')" tooltip="View source in MATLAB Online">Open in MATLAB Online</button>
+            <t:help href="https://github.com/jenkinsci/matlab-plugin/blob/master/CONFIGDOC.md" tooltip="MATLAB Plugin Configuration Guide" />
```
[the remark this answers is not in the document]
mw-kapilg: I think it would be valuable to keep "MATLAB" in the heading as these are just MATLAB test results, the user might be running other tests in their job whose results we won't be showing here.

src/main/resources/com/mathworks/ci/TestResultsViewAction/index.jelly
mw-kapilg: Since colon is something already associated with showing time (like HH:MM:SS) in some sense, to avoid confusion I chose =

src/main/java/com/mathworks/ci/TestCase.java
```diff
@@ -0,0 +1,105 @@
...
+    private List<TestDiagnostics> diagnostics;
+    private boolean passed;
+    private boolean failed;
+    private boolean incomplete;
+    private String status;
+    private Double duration;
+    private String id;
+
+    public TestCase() {
+        name = "";
```
nbhoski, line 24: Its good practice to use this for class properties

src/main/java/com/mathworks/ci/TestDiagnostics.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.apache.commons.lang.RandomStringUtils;
+
+public class TestDiagnostics {
+    private String event;
+    private String report;
+    private String id;
+
+    public TestDiagnostics() {
+        event = "";
```
nbhoski, line 16: consider using `this` for property

src/main/java/com/mathworks/ci/TestFile.java
```diff
@@ -0,0 +1,94 @@
...
+    private List<TestCase> testCases;
+    private String id;
+
+    public TestFile() {
+        filePath = "";
+        name = "";
+        duration = 0.0;
+        status = "NotRun";
+        testCases = new ArrayList<TestCase>();
+
```
nbhoski, line 27: Extra space

src/main/java/com/mathworks/ci/TestFile.java
```diff
@@ -0,0 +1,94 @@
...
+        id = RandomStringUtils.randomAlphanumeric(8);
+    }
+
+    public void incrementDuration(Double testCaseDuration) {
+        this.duration += testCaseDuration;
+    }
+
+    public void updateStatus(TestCase testCase) {
+        if (!status.equals("Failed")){
```
nbhoski, line 36: Although its hardcoded its good to check it with equalsIgnoreCase() for future changes from Failed to failed

src/main/java/com/mathworks/ci/TestFile.java
```diff
@@ -0,0 +1,94 @@
...
+            if (testCase.getFailed()){
+                status = "Failed";
+            }
+            else if (!status.equals("Incomplete")){
```
nbhoski, line 40: Same as above

src/main/java/com/mathworks/ci/TestFile.java
```diff
@@ -0,0 +1,94 @@
...
+                }
+            }
+        }
+    }
+
+    public String getFilePath() {
+        return this.filePath;
+    }
+
+    public void setFilePath(String filePath) {
```
nbhoski, line 55: I would refrain using `filePath` as name just to keep the class name differentiated as we heavily use FilePath class. variable may cause confusion.

src/main/resources/com/mathworks/ci/TestResultsViewAction/index.jelly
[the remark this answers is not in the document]
mw-hrastega: I think "MATLAB Test Results" is fine. My comment was about "MATLAB Plugin Configuration Guide."

src/main/resources/com/mathworks/ci/TestResultsViewAction/index.jelly
mw-hrastega: I think a single colon would work; however, I agree that something like (HH:MM:SS): would be confusing. The equals sign might be misinterpreted as an assignment operator :)

src/test/java/integ/com/mathworks/ci/TestResultsViewActionTest.java
```diff
@@ -0,0 +1,427 @@
...
+    public void verifyIncompleteTestsCount() throws ExecutionException, InterruptedException, URISyntaxException, IOException, ParseException {
+        FreeStyleBuild build = getFreestyleBuild();
+        final FilePath workspace = new FilePath(new File("C:", "workspace"));
+        final String actionID = "abc123";
+        final String targetFile = MatlabBuilderConstants.TEST_RESULTS_VIEW_ARTIFACT + actionID + ".json";
+        FilePath artifactRoot = new FilePath(build.getRootDir());
+        copyFileInWorkspace("testArtifacts/t1/windows/" + MatlabBuilderConstants.TEST_RESULTS_VIEW_ARTIFACT + ".json",targetFile,artifactRoot);
+        TestResultsViewAction ac = new TestResultsViewAction(build, workspace, actionID);
+        int actualCount = ac.getIncompleteCount();
+        Assert.assertEquals("Incorrect incomplete tests count",0,actualCount);
```
sameagen-MW, line 168: We should maybe test with a non-default value instead of 0?

src/main/resources/com/mathworks/ci/TestResultsViewAction/index.jelly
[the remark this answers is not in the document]
mw-kapilg: Sure, that makes sense! Updated

src/main/resources/com/mathworks/ci/TestResultsViewAction/index.jelly
mw-kapilg: Good point, updated!

src/main/resources/com/mathworks/ci/TestResultsViewAction/index.jelly
mw-kapilg: Updated (seconds) to (s)

> Is it better to include + and - signs or drop-down arrows instead?

I didn't get this, could you explain?

src/main/resources/com/mathworks/ci/TestResultsViewAction/index.jelly
mw-kapilg: Sure,  if that doesn't cause any confusion!

src/main/java/com/mathworks/ci/TestCase.java
```diff
@@ -0,0 +1,105 @@
...
+ * Copyright 2024, The MathWorks Inc.
+ *
+ */
+
+import java.util.List;
+import java.util.ArrayList;
+
+import org.apache.commons.lang.RandomStringUtils;
+
+public class TestCase {
```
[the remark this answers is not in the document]
mw-kapilg: Updated class names for then `TestFile`, `TestCase` and `TestDiagnostics`

src/main/java/com/mathworks/ci/TestCase.java
```diff
@@ -0,0 +1,105 @@
...
+    private Double duration;
+    private String id;
+
+    public TestCase() {
+        name = "";
+        diagnostics = new ArrayList<TestDiagnostics>();
+        passed = false;
+        failed = false;
+        incomplete = false;
+        status = "NotRun";
```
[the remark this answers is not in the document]
mw-kapilg: Added status constants in `MatlabBuilderConstants` and updated instances to refer to those

src/main/java/com/mathworks/ci/TestFile.java
mw-kapilg: Using static constants now

src/main/java/com/mathworks/ci/TestFile.java
mw-kapilg: Yeah, I agree. Updated `filePath` to just `path`

src/main/java/com/mathworks/ci/TestCase.java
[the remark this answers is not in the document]
mw-kapilg: I've added status constants in `MatlabBuilderConstants` instead. Felt it wouldn't be drastically different, though let me know if you have a flag against this

src/main/java/com/mathworks/ci/TestResultsViewAction.java
```diff
@@ -0,0 +1,265 @@
...
+        passedCount = 0;
+        failedCount = 0;
+        incompleteCount = 0;
+        notRunCount = 0;
+
+         try{
+            // Set test results counts
+             getTestResults();
+         } catch (InterruptedException | IOException e) {
+             e.printStackTrace();
```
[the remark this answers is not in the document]
mw-kapilg: Not really, removed it

src/main/java/com/mathworks/ci/TestResultsViewAction.java
```diff
@@ -0,0 +1,265 @@
...
+
+                if(jsonTestSessionResults instanceof JSONArray){
+                    JSONArray jsonTestSessionResultsArray = (JSONArray) jsonTestSessionResults;
+                    Iterator<JSONObject> testSessionResultsIterator = jsonTestSessionResultsArray.iterator();
+
+                    while(testSessionResultsIterator.hasNext()){
+                        JSONObject jsonTestCase = testSessionResultsIterator.next();
+                        getTestSessionResults(testSessionResults, jsonTestCase, map);
+                    }
+                }
```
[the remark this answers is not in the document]
mw-kapilg: I can't see it 👀

src/main/java/com/mathworks/ci/TestResultsViewAction.java
```diff
@@ -0,0 +1,265 @@
...
+                else if(jsonTestSessionResults instanceof JSONObject) {
```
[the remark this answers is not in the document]
mw-kapilg: I understand it's not super intuitive. If you have a look at the [artifact](https://github.com/mathworks/jenkins-matlab-plugin/pull/360/files#diff-0454a99a615b34671023fdd246f8a948e833dad50f22f2143a90108e4de507de) we use for our unit tests, you'll see that inside the top level `JSONArray`, we have a `JSONArray` followed by a `JSONObject`. This mainly happens because of how we handle multiple test sessions' result and json encode them. Only in the case where a test session has a single test which ran, we'll have an instance of `JSONObject`. Hope this helps, feel free to reach out otherwise!

src/main/java/com/mathworks/ci/TestResultsViewAction.java
```diff
@@ -0,0 +1,265 @@
...
+        }
+        catch (Exception e) {
+            throw new IOException(e.getLocalizedMessage());
+        }
+
+        return testResults;
+    }
+
+    private void getTestSessionResults(List<TestFile> testSessionResults, JSONObject jsonTestCase, Map<String, TestFile> map) throws IOException, InterruptedException {
+        FilePath baseFolder = new FilePath(new File(jsonTestCase.get("BaseFolder").toString()));
```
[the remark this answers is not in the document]
mw-kapilg: Yes, it's absolute path

src/test/java/integ/com/mathworks/ci/TestResultsViewActionTest.java
mw-kapilg: Thanks for catching that, updated the test!

src/main/java/com/mathworks/ci/TestResultsViewAction.java
```diff
@@ -0,0 +1,265 @@
...
+        String testFileName = testNameSplit[0];
+        String testCaseName = testNameSplit[1];
+
+        TestFile testFile = map.get(baseFolder + File.separator + testFileName);
+        if(testFile == null) {
+            testFile = new TestFile();
+            testFile.setName(testFileName);
+
+            map.put(baseFolder + File.separator + testFileName, testFile);
+            testSessionResults.add(testFile);
```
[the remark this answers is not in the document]
mw-kapilg: If a test file is in multiple sessions, it will get added to all respective `testSessionResults` arrays. At line 80, you can see that the `testSessionResults` array gets initialised at the beginning of when we start reading results of a particular test session.
This `if` block makes sure within a `testSessionResults` array, we have exactly one instance of each test file that ran in the particular test session. Multiple test cases of the same test file are stored within the `MatlabTestFile` object as a list of `MatlabTestCase`.

src/main/java/com/mathworks/ci/TestResultsViewAction.java
```diff
@@ -0,0 +1,265 @@
...
+        catch (Exception e) {
+            throw new IOException(e.getLocalizedMessage());
+        }
+
+        return testResults;
+    }
+
+    private void getTestSessionResults(List<TestFile> testSessionResults, JSONObject jsonTestCase, Map<String, TestFile> map) throws IOException, InterruptedException {
+        FilePath baseFolder = new FilePath(new File(jsonTestCase.get("BaseFolder").toString()));
+        JSONObject testCaseResult = (JSONObject) jsonTestCase.get("TestResult");
```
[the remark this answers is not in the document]
mw-kapilg: It sure can if the `JSONObject` doesn't have the keys `"BaseFolder"` or `"TestResult"` but don't know if & when it would happen

src/main/java/com/mathworks/ci/TestResultsViewAction.java
```diff
@@ -0,0 +1,265 @@
...
+             e.printStackTrace();
+             throw e;
+         } catch (ParseException e) {
+             e.printStackTrace();
+         }
+    }
+
+    public List<List<TestFile>> getTestResults() throws ParseException, InterruptedException, IOException {
+        List<List<TestFile>> testResults = new ArrayList<>();
+        FilePath fl = new FilePath(new File(build.getRootDir().getAbsolutePath() + File.separator + MatlabBuilderConstants.TEST_RESULTS_VIEW_ARTIFACT + this.actionID + ".json"));
```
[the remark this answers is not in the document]
mw-kapilg: Don't get me started on file paths issue with Jenkins, I've devoted so much time to fixing those 😭  
Haven't tested this combination yet but sure will test it before releasing.

src/main/resources/com/mathworks/ci/TestResultsViewAction/index.jelly
mw-hrastega: Nothing major. Currently, it seems that you use the + and - buttons to expand and collapse sections in the report. I was wondering whether a drop-down arrow could do the same. In the test browser, we resort to the arrow for expansion: https://www.mathworks.com/help/matlab/matlab_prog/run-tests-using-test-browser.html#:~:text=You%20can%20also%20add%20tests%20to%20the%20test%20browser%20by%20clicking%20the%20drop%2Ddown%20arrow

src/main/resources/com/mathworks/ci/TestResultsViewAction/index.jelly
```diff
@@ -0,0 +1,217 @@
...
+                                </td>
+                            </tr>
+                        </j:forEach>
+                    </tbody>
+                </table>
+            </j:forEach>
+        </l:main-panel>
+    </l:layout>
+    <script type="text/javascript">
+        function toggleVisibilty(id) {
```
sameagen-MW, line 153: I remember there being some weird permission stuff with JS here. Does the user still need to manually jump through some hoops to enable this?

src/test/java/integ/com/mathworks/ci/TestResultsViewActionTest.java
```diff
@@ -0,0 +1,433 @@
...
+    public void verifyIncompleteTestsCount() throws ExecutionException, InterruptedException, URISyntaxException, IOException, ParseException {
+        FreeStyleBuild build = getFreestyleBuild();
+        final FilePath workspace = new FilePath(new File("C:", "workspace"));
+        final String actionID = "abc123";
+        final String targetFile = MatlabBuilderConstants.TEST_RESULTS_VIEW_ARTIFACT + actionID + ".json";
+        FilePath artifactRoot = new FilePath(build.getRootDir());
+        copyFileInWorkspace("testArtifacts/t1/windows/" + MatlabBuilderConstants.TEST_RESULTS_VIEW_ARTIFACT + ".json",targetFile,artifactRoot);
+        TestResultsViewAction ac = new TestResultsViewAction(build, workspace, actionID);
+        int actualCount = ac.getIncompleteCount();
+        Assert.assertEquals("Incorrect incomplete tests count",1,actualCount);
```
sameagen-MW, line 167: Using different numbers for each of these rather than 1 for all could catch weird cases where one field was incremented by the wrong thing. Something like 3 passed, 2 failed, and 1 incomplete to uniquify the values. Pretty edge case.

src/main/resources/+ciplugins/+jenkins/TestResultsViewPlugin.m
```diff
@@ -0,0 +1,33 @@
+classdef TestResultsViewPlugin < matlab.unittest.plugins.TestRunnerPlugin
+    % Copyright 2024 The MathWorks, Inc.
+    
+    methods (Access=protected)
+        function reportFinalizedSuite(plugin, pluginData)
+            % Checkout MATLAB Test license
+            license('checkout', 'matlab_test');
```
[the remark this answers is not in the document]
davidbuzinski: From the docs, it seems like it returns a `status` and `errmsg`, but I don't think it throws. The status will just be a boolean so I'm wondering if it might continue on even if the user doesn't have MATLAB Test.

Also I'm thinking even if we short circuit like that we might need to be sure to invoke the super class method so this method runs on all the plugins.

src/main/java/com/mathworks/ci/MatlabTestCase.java
```diff
@@ -0,0 +1,107 @@
...
+        if (this.failed){
+            this.status = MatlabBuilderConstants.FAILED;
+        }
+        else if (this.incomplete) {
+            this.status = MatlabBuilderConstants.INCOMPLETE;
+        }
+        else if(this.passed) {
+            this.status = MatlabBuilderConstants.PASSED;
+        }
+    }
```
davidbuzinski, line 46: This model of having booleans for `passed`, `failed`, and `incomplete` (which I think might be mutually exclusive in our implementation here but I'm not totally sure) along with a `status` property that seems to have all of that same data wrapped up into one field seems strange to me.

It looks like this `updateStatus` gets called after we set all those 3 other properties, and is derived from the same information, but I'm not sure why we can't just have just one `status` enum property or the booleans.

src/main/java/com/mathworks/ci/MatlabTestFile.java
```diff
@@ -0,0 +1,95 @@
...
+            else if (!this.status.equals(MatlabBuilderConstants.INCOMPLETE)){
+                if (matlabTestCase.getIncomplete()){
+                    this.status = MatlabBuilderConstants.INCOMPLETE;
+                }
+                else if (matlabTestCase.getPassed()){
+                    this.status = MatlabBuilderConstants.PASSED;
+                }
+            }
+        }
+    }
```
davidbuzinski, line 50: Seems a little strange here too with the `updateStatus` method because it only uses the boolean properties of the MatlabTestCase and not the status. It looks like we update the MatlabTestFile status as each MatlabTestCase comes in.

src/main/java/com/mathworks/ci/TestResultsViewAction.java
```diff
@@ -0,0 +1,265 @@
...
+        String matlabTestFileName = testNameSplit[0];
+        String matlabTestCaseName = testNameSplit[1];
+
+        // Check if test's file was known or not
+        MatlabTestFile matlabTestFile = map.get(baseFolder + File.separator + matlabTestFileName);
+        if(matlabTestFile == null) {
+            matlabTestFile = new MatlabTestFile();
+            matlabTestFile.setName(matlabTestFileName);
+
+            map.put(baseFolder + File.separator + matlabTestFileName, matlabTestFile);
```
davidbuzinski, line 122: Does `File.separator` work if the Jenkins server is running on Windows but the job is run on a Linux node? I was seeing some weirdness with that case with the mpm implementation, but I'm not sure if all of this code runs on the server or if part is running on the node as well. We might want to make sure we bash that case.

src/main/java/com/mathworks/ci/TestResultsViewAction.java
```diff
@@ -0,0 +1,265 @@
...
+                matlabTestDiagnostics.setEvent(diagnosticItem.get("Event").toString());
+                matlabTestDiagnostics.setReport(diagnosticItem.get("Report").toString());
+                matlabTestCase.getDiagnostics().add(matlabTestDiagnostics);
+            }
+        }
+
+        matlabTestFile.incrementDuration(matlabTestCase.getDuration());
+        matlabTestFile.updateStatus(matlabTestCase);
+        matlabTestFile.getMatlabTestCases().add(matlabTestCase);
+        updateCount(matlabTestCase);
```
davidbuzinski, line 167: Maybe we can have an `addTestCase` method on the MatlabTestFile that handles these responsibilities.

src/main/java/com/mathworks/ci/TestResultsViewAction.java
```diff
@@ -0,0 +1,265 @@
...
+        Path path2 = Paths.get(this.workspace.toURI());
+        Path relPath = path2.relativize(path1);
+
+        matlabTestFile.setPath(this.workspace.getName() + File.separator + relPath.toString());
+
+        MatlabTestCase matlabTestCase = new MatlabTestCase();
+        matlabTestCase.setName(matlabTestCaseName);
+        matlabTestCase.setPassed((boolean) matlabTestCaseResult.get("Passed"));
+        matlabTestCase.setFailed((boolean) matlabTestCaseResult.get("Failed"));
+        matlabTestCase.setIncomplete((boolean) matlabTestCaseResult.get("Incomplete"));
```
davidbuzinski, line 137: Maybe the `MatlabTestCase` constructor can take the name as well (as Sam suggested above for `MatlabTestFile`). I am wondering if we can just derive and set the status directly here and get rid of those boolean properties on `MatlabTestCase` or if we're also using them somewhere else.

davidbuzinski: Something like
```
if (matlabTestCaseResult.get("Passed")) {
    matlabTestCase.setStatus(MatlabBuilderConstants.PASSED)
} else if (matlabTestCaseResult.get("Failed")) {
    matlabTestCase.setStatus(MatlabBuilderConstants.FAILED)
} else if (matlabTestCaseResult.get("Incomplete")) {
    matlabTestCase.setStatus(MatlabBuilderConstants.INCOMPLETE)
}
```
It's a little messier here but it's pretty much the same as what's in the `updateStatus` method.

src/main/java/com/mathworks/ci/TestResultsViewAction.java
```diff
@@ -0,0 +1,265 @@
...
+        // Not OS dependent
+        String[] testNameSplit = matlabTestCaseResult.get("Name").toString().split("/");
+        String matlabTestFileName = testNameSplit[0];
+        String matlabTestCaseName = testNameSplit[1];
+
+        // Check if test's file was known or not
+        MatlabTestFile matlabTestFile = map.get(baseFolder + File.separator + matlabTestFileName);
+        if(matlabTestFile == null) {
+            matlabTestFile = new MatlabTestFile();
+            matlabTestFile.setName(matlabTestFileName);
```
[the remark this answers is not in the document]
mw-kapilg: Good idea, updated the constructor

src/main/resources/com/mathworks/ci/TestResultsViewAction/index.jelly
mw-kapilg: Nope, not for basic JavaScript

src/test/java/integ/com/mathworks/ci/TestResultsViewActionTest.java
mw-kapilg: Updated, thanks!

src/main/java/com/mathworks/ci/MatlabTestCase.java
mw-kapilg: You're actually correct, the status was needed in a previous view in which the info on both the overall test case status as well as Boolean values for each status type was required (for e.g. in case of an assertion failure, the overall test case status would be `failed` but that test case is also marked as `incomplete`). Since we no longer need it, I've removed the Booleans and introduced a `TestStatus` enum.

src/main/java/com/mathworks/ci/TestResultsViewAction.java
mw-kapilg: Done, thanks!

src/main/java/com/mathworks/ci/TestResultsViewAction.java
mw-kapilg: Updated the constructor, thanks! Rest of the part is addressed in the above comment

src/main/java/com/mathworks/ci/TestResultsViewAction.java
mw-kapilg: I remember Sam had mentioned it earlier, since all the action code would only be running on the agent we shouldn't face issues there. But it would be good to bash this scenario, thanks for reminding!

src/main/java/com/mathworks/ci/TestResultsViewAction.java
```diff
@@ -0,0 +1,265 @@
...
+        this.incompleteCount = 0;
+        this.notRunCount = 0;
+
+         try{
+            // Set test results counts
+             getTestResults();
+         } catch (InterruptedException | IOException e) {
+             throw e;
+         } catch (ParseException e) {
+             e.printStackTrace();
```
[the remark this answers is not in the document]
mw-kapilg: Yes, when there might be a parsing error with the JSON file the test results won't appear in Jenkins. I don't think there should be any other consequences

src/main/resources/+ciplugins/+jenkins/TestResultsViewPlugin.m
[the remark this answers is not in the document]
mw-kapilg: To add `TestResultsViewPlugin` to the test runner, we use a [`TestResultsViewPluginService`](https://github.com/mathworks/jenkins-matlab-plugin/pull/360/files#diff-a5d725ac796e54f2ca64056da44fd9a5cc5baf54549401d2c999e778aa49f826R7). The service includes a check to test whether the user has a valid MATLAB Test license, if so only then does the plugin get added to the test runner otherwise not.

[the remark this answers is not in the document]
davidbuzinski: I see, makes sense. So all of the license checking should have already happened by this point, but this line is actually checking out the license so that it shows up as in use for this session. Thanks for clearing that up!

src/main/resources/com/mathworks/ci/TestResultsViewAction/index.jelly
```diff
@@ -0,0 +1,217 @@
...
+            <button id="Failed" onclick="showTests(id)" class="jenkins-button jenkins-!-margin-4 jenkins-!-padding-0" style="display: inline; width:8em;" tooltip="Failed tests">
+                <p>${it.failedCount}<br>
+                    <l:icon class="symbol-close-circle-outline plugin-ionicons-api icon-lg jenkins-!-error-color"/>
+                </br><br></br>Failed</p>
+            </button>
+
+            <button id="Incomplete" onclick="showTests(id)" class="jenkins-button jenkins-!-margin-4 jenkins-!-padding-0" style="display: inline; width:8em;" tooltip="Incomplete tests">
+                <p>${it.incompleteCount}<br>
+                    <l:icon class="symbol-alert-circle-outline plugin-ionicons-api icon-lg jenkins-!-warning-color"/>
+                </br><br></br>Incomplete</p>
```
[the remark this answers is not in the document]
mw-kapilg: Yeah, these are closest I could find from icons available here: https://ionic.io/ionicons. Lemme know if you think some other icon would suit better

src/main/resources/com/mathworks/ci/TestResultsViewAction/index.jelly
mw-kapilg: That's true. For reference I had a look at the Jenkins JUnit and TAP plugins which were using + and - symbols to expand and collapse. I don't mind the chevrons either, do you feel using them for dropdowns would be more helpful for the user?

src/main/resources/+ciplugins/+jenkins/TestResultsViewPlugin.m
```diff
@@ -0,0 +1,33 @@
...
+                testResults = {jsondecode(fileread(testArtifactFile))};
+            else
+                testResults = {};
+            end
+            testResults{end+1} = testDetails;
+            JsonTestResults = jsonencode(testResults, "PrettyPrint", true);
+            
+            fID = fopen(testArtifactFile, "w");
+            fprintf(fID, '%s', JsonTestResults);
+            fclose(fID);
```
davidbuzinski, line 27: Check how we handle writing to the file in the `BuildReportPlugin` class. We issue a warning if `fopen` returns `-1`, and clean up a little differently using `onCleanup` (which is safer). I think this was the source of a bug when we first introduced the build report to the Jenkins plugin. Otherwise it might fail the test session/ build because of a weird filesystem permissions issue.

src/main/java/com/mathworks/ci/TestResultsViewAction.java
```diff
@@ -0,0 +1,270 @@
...
+
+    public List<List<MatlabTestFile>> getTestResults() throws ParseException, InterruptedException, IOException {
+        List<List<MatlabTestFile>> testResults = new ArrayList<>();
+        FilePath fl = new FilePath(new File(build.getRootDir().getAbsolutePath(), MatlabBuilderConstants.TEST_RESULTS_VIEW_ARTIFACT + this.actionID + ".json"));
+        try (InputStreamReader reader = new InputStreamReader(Files.newInputStream(Paths.get(fl.toURI())), StandardCharsets.UTF_8)) {
+            this.totalCount = 0;
+            this.passedCount = 0;
+            this.failedCount = 0;
+            this.incompleteCount = 0;
+            this.notRunCount = totalCount;
```
sameagen-MW, line 80: Maybe clearer (and less accident prone when we make changes later) to just use `this.notRunCount = 0`

src/main/java/com/mathworks/ci/TestResultsViewAction.java
```diff
@@ -0,0 +1,270 @@
...
+            matlabTestFile = new MatlabTestFile(matlabTestFileName);
+
+            map.put(baseFolder + File.separator + matlabTestFileName, matlabTestFile);
+            testSessionResults.add(matlabTestFile);
+        }
+
+        // Calculate the relative path
+        Path path1 = Paths.get(baseFolder.toURI());
+        Path path2 = Paths.get(this.workspace.toURI());
+        Path relPath = path2.relativize(path1);
```
sameagen-MW, line 135: We can maybe use clearer variable names here?

src/main/resources/+ciplugins/+jenkins/TestResultsViewPlugin.m
```diff
@@ -0,0 +1,33 @@
+classdef TestResultsViewPlugin < matlab.unittest.plugins.TestRunnerPlugin
+    % Copyright 2025 The MathWorks, Inc.
+    
+    methods (Access=protected)
+        function reportFinalizedSuite(plugin, pluginData)
+            % Checkout MATLAB Test license
+            license('checkout', 'matlab_test');
```
sameagen-MW, line 7: I'm still slightly worried about this. I think ideally for me we'd do the license check and checkout at the same time, just to avoid any situations where MATLAB Test was available initially at the time of the first check but not at the time of reporting the finalized suite. Probably not a big deal though.

sameagen-MW (approved):
Just some minor comments from me! The only blocker I see is what David mentioned with the cleanup code for the plugin. 

:shipit:

src/main/resources/+ciplugins/+jenkins/TestResultsViewPlugin.m
```diff
@@ -0,0 +1,33 @@
...
+    methods (Access=protected)
+        function reportFinalizedSuite(plugin, pluginData)
+            % Checkout MATLAB Test license
+            license('checkout', 'matlab_test');
+
+            testDetails = {};
+            for idx = 1:numel(pluginData.TestResult)
+                testDetails(idx).TestResult = pluginData.TestResult(idx);
+                testDetails(idx).BaseFolder = pluginData.TestSuite(idx).BaseFolder;
+            end
```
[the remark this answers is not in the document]
mw-kapilg: Yeah, I got a little confused when json encoding/decoding with struct, cells from MATLAB. You're right we can declare `testDetails` as an empty struct.

src/main/resources/+ciplugins/+jenkins/TestResultsViewPlugin.m
mw-kapilg: Updated, thanks!

src/main/resources/+ciplugins/+jenkins/TestResultsViewPlugin.m
mw-kapilg: I think it should be fine. Do you have a case in mind where the situation you're saying could happen?

davidbuzinski (approved):

PR #361, opened 2024-09-25
documenting changes to the table of task results

mw-hrastega:
Hi, Kapil. I decided not to explicitly mention the Run MATLAB Command/`runMATLABCommand` step for this feature. The existing text sounds comprehensive enough.

Once we finalize this pull request, I will need to get a quick editorial review for the added sentence.

### DIFF
--- a/CONFIGDOC.md
+++ b/CONFIGDOC.md
@@ -69,9 +69,9 @@ You can specify build options for your MATLAB build by first selecting **Build o
 
 MATLAB exits with exit code 0 if the specified tasks run without error. Otherwise, MATLAB terminates with a nonzero exit code, which causes the step to fail.
 
-Starting in R2024a, you can access the result of running a MATLAB build interactively in your Jenkins interface. After your build runs, the Jenkins build summary page displays the number of tasks that ran, failed, and were skipped. You can click the **MATLAB Build Results** link on the page to access the table of task results. The table provides information about each task that was part of the MATLAB build. Clicking a task name in the table displays relevant build log information on the **Console Output** page.
+Starting in R2024a, you can access the result of running a MATLAB build interactively in your Jenkins interface. After your build runs, the Jenkins build summary page displays the number of tasks that ran, failed, and were skipped. You can click the **MATLAB Build Results** link on the page to access the table of task results. The table provides information about each task that was part of the MATLAB build. For example, if a task was skipped, the table displays the reason that the task was skipped. Clicking a task name in the table displays relevant build log information on the **Console Output** page.
 
-![matlab_build_results](https://github.com/mathworks/jenkins-matlab-plugin/assets/48831250/bf3a8f01-12d6-436e-8a9e-10f465a98f6a)
+![Table of MATLAB build results including three tasks](https://github.com/user-attachments/assets/eaef15b8-05d3-49c4-ae85-4aca1e57e324)
 
 #### Run MATLAB Tests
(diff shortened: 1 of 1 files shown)


CONFIGDOC.md
```diff
@@ -69,9 +69,9 @@ You can specify build options for your MATLAB build by first selecting **Build o
...
 
-Starting in R2024a, you can access the result of running a MATLAB build interactively in your Jenkins interface. After your build runs, the Jenkins build summary page displays the number of tasks that ran, failed, and were skipped. You can click the **MATLAB Build Results** link on the page to access the table of task results. The table provides information about each task that was part of the MATLAB build. Clicking a task name in the table displays relevant build log information on the **Console Output** page.
+Starting in R2024a, you can access the result of running a MATLAB build interactively in your Jenkins interface. After your build runs, the Jenkins build summary page displays the number of tasks that ran, failed, and were skipped. You can click the **MATLAB Build Results** link on the page to access the table of task results. The table provides information about each task that was part of the MATLAB build. For example, if a task was skipped, the table displays the reason that the task was skipped. Clicking a task name in the table displays relevant build log information on the **Console Output** page.
```
sameagen-MW, line 72: Could we change "access the result" -> "browse the results"? Or just change "result" to "results"?

The skip message seems like a somewhat niche example if we're just trying to list examples of the type of information found in the table since it doesn't apply to every task. I would expect the user to care more about name, status, and duration. Is there a secondary reason that we're pointing out the skip reason specifically?

sameagen-MW, line 72: "Clicking a task name in the table displays relevant build log information on the **Console Output** page." -> "Clicking a task name in the table navigates to the relevant section of the build log on the **Console Output** page" 

to make it clear that we're actually just changing pages rather than pulling information from that location onto the current page?

mw-hrastega: The current text doesn't sound confusing. I will run this comment by the editor and let her decide.

mw-hrastega: What do you mean by 'Or just change "result" to "results"?'

sameagen-MW: Ah gotcha. I found it a bit confusing but not a big deal

sameagen-MW: > What do you mean by 'Or just change "result" to "results"?'

We switch from referring to a single result in the first sentence to using plural results in the third sentence. I think that using the plural "results" everywhere feels a little bit smoother to read. And I just thought that "browse" was a slightly more descriptive verb than "access" here. I don't normally think of myself as "accessing" results, it's more of a "browsing results", "viewing results", "checking results", etc...

mw-kapilg: > I added the skip example because the "skip reason" sounded like an important aspect of this minor release. However, I am happy to remove the added sentence.

I am fine if we don't want to keep it since the updated screenshot shows skip reason now.

> We switch from referring to a single result in the first sentence to using plural results in the third sentence. I think that using the plural "results" everywhere feels a little bit smoother to read.

Agree, I prefer using results everywhere for consistency

mw-kapilg (approved):
Looks good, thanks Houman!

CONFIGDOC.md
mw-hrastega: > think that using the plural "results" everywhere feels a little bit smoother to read. And I just thought that "browse" was a slightly more descriptive verb than "access" here. I don't normally think of myself as "accessing" results, it's more of a "browsing results", "viewing results", "checking results", etc...

I changed result to results and removed the skip-related sentence. Regarding the best verb to use here, I will defer to the editor. Thanks Sam!

PR #366, opened 2024-11-05
Use ToolInstaller

davidbuzinski:
* Updated MatlabInstaller to inherit from ToolInstaller rather than DownloadFromUrlInstaller. This removed the need for a MatlabInstallable class
* Updated prereleases to use `--release-status` flag (ended up being resolved in the parent branch anyways)
* Removed spaces between method names and open parenthesis to be consistent with rest of code base. Open to ideas about updating it elsewhere, but we should be consistent (and maybe add linter rules)

Still to do/ to consider:
* Change the word version to release everywhere (like in other other CI offerings and the mpm API)
* Update the mpmInstall method to have an API more similar to mpm


nbhoski:
Thaks @davidbuzinski for these changes. these looks good. and i support of these changes. My rationale was to maintain the single object responsibility by separating `MatlabInstallable` class . also as mentioned in one of my older comments we are still downloading our tool before we perform the installation which also supports for platform specific `MatlabInstallable` implementation.

nbhoski (approved):
LGTM!

davidbuzinski:
> Thaks @davidbuzinski for these changes. these looks good. and i support of these changes. My rationale was to maintain the single object responsibility by separating `MatlabInstallable` class . also as mentioned in one of my older comments we are still downloading our tool before we perform the installation which also supports for platform specific `MatlabInstallable` implementation.
> 
> I have added following in this branch
> 
>     * Change the word version to release everywhere (like in other other CI offerings and the mpm API)
> 
>     * Update the mpmInstall method to have an API more similar to mpm

Thanks for making those changes, they look great

PR #375, opened 2024-11-13
Remove unused imports and fix validation bug

davidbuzinski:
The release validation had an annotation that was causing some sort of http request. It ended up causing a weird bug during form validation.

Screenshot 2024-11-13 130250


sameagen-MW:
I think that all form validation is done via HTTP, but it seems like that annotation forces that 404 if not using a POST request: https://www.jenkins.io/doc/developer/security/form-validation/#protecting-from-csrf. Since we're not doing anything with sideeffects I think removing it will be an improvement!

sameagen-MW (approved):

PR #377, opened 2024-11-14
remove unused imports and fix validation bug (#375)

davidbuzinski:
Updating qual branch with changes from https://github.com/mathworks/jenkins-matlab-plugin/pull/375


nbhoski (approved):
This is already part of qual branch however import formats would add in.

PR #382, opened 2024-12-06
Update to standard formatting

davidbuzinski:
Just applied one of the default Java formatters from VS Code to all of the files and updated the copyrights accordingly. No functional changes


sameagen-MW (approved):
:shipit:

PR #391, opened 2025-01-02
Fix matlab batch path bug

davidbuzinski:
This change puts `matlab-batch` in `$MATLAB_ROOT/bin` directory and adding that to path. The reason for this is because in our current `MatlabInstallation` code there is no info about the node or node channel being run. If we use FilePath as I before, there is a problem where it will use Windows style paths on Linux nodes if the server is hosted on Windows.

It is also a bit strange that we add `$MATLAB_ROOT/..` (installation home) to path even for MATLAB installations that were not configured by Install Automatically. This could add unintended folders to path for self-managed MATLAB installations.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
 	</parent>
 
 	<artifactId>matlab</artifactId>
-	<version>2.15.1-SNAPSHOT</version>
+	<version>2.16.0-SNAPSHOT</version>
 	<packaging>hpi</packaging>
 
 	<name>MATLAB Plugin</name>
--- a/src/main/java/com/mathworks/ci/MatlabInstallation.java
+++ b/src/main/java/com/mathworks/ci/MatlabInstallation.java
@@ -64,13 +64,7 @@ public class MatlabInstallation extends ToolInstallation
         if (home == null) {
             return;
         }
-        FilePath matlabHome = new FilePath(new File(home));
-        FilePath matlabBin = new FilePath(matlabHome, "bin");
-        env.put("PATH+matlabroot", matlabBin.getRemote());
-        FilePath matlabBatchFolder = matlabHome.getParent();
-        if (matlabBatchFolder != null) {
-            env.put("PATH+matlabbatch", matlabBatchFolder.getRemote());
-        }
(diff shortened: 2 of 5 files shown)


src/main/java/com/mathworks/ci/MatlabInstallation.java
```diff
@@ -64,13 +64,7 @@ public void buildEnvVars(EnvVars env) {
...
             return;
         }
-        FilePath matlabHome = new FilePath(new File(home));
-        FilePath matlabBin = new FilePath(matlabHome, "bin");
-        env.put("PATH+matlabroot", matlabBin.getRemote());
-        FilePath matlabBatchFolder = matlabHome.getParent();
-        if (matlabBatchFolder != null) {
-            env.put("PATH+matlabbatch", matlabBatchFolder.getRemote());
-        }
+        env.put("PATH+matlabroot", home + "/bin");
```
sameagen-MW, line 67: Does this work on Windows with the hardcoded forward slash?

davidbuzinski: It seems that it does. This is close to the original implementation before the Install Automatically changes

davidbuzinski: We want to specify a remote path on the node running the job, but `File` and `FilePath` both use the path separator of the OS of the server unless you use the constructor for `FilePath` where we specify a VirtualChannel for the node. In `Utilities` and `UseMatlabVersionBuildWrapper` we have access to the `Node` object or the `Launcher` object that can give us a `VirtualChannel`, but in `MatlabInstallation` we aren't keeping that info for access in this method.

pom.xml
```diff
@@ -11,7 +11,7 @@
 	</parent>
 
 	<artifactId>matlab</artifactId>
-	<version>2.15.1-SNAPSHOT</version>
+	<version>2.16.0-SNAPSHOT</version>
```
Vahila, line 14: Why is the version updated? Generally, the version is updated during the maven release process.

sameagen-MW (approved):

Issue #392, opened 2025-01-03
Using 'latest' release installs MATLAB under '<Tool location>/latest' directory instead of '<Tool location>/R202xx'


Vahila:
Using 'latest' release installs MATLAB under '<Tool location>/latest' directory instead of '<Tool location>/R202xx'. This would cause an error when the latest points to the new release.

PR #393, opened 2025-01-07
Resolve latest release issue

davidbuzinski:
Resolve latest release and use that to determine name of MATLAB folder. Also revert change where I bumped version in `pom.xml` because Vahila said that's part of the release process.

https://github.com/mathworks/jenkins-matlab-plugin/issues/392

### DIFF
--- a/src/main/java/com/mathworks/ci/tools/MatlabInstaller.java
+++ b/src/main/java/com/mathworks/ci/tools/MatlabInstaller.java
@@ -1,7 +1,7 @@
 package com.mathworks.ci.tools;
 
 /**
- * Copyright 2024, The MathWorks, Inc.
+ * Copyright 2024-2025, The MathWorks, Inc.
  */
 
 import com.mathworks.ci.MatlabInstallation;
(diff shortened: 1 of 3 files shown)


src/main/java/com/mathworks/ci/tools/MatlabInstaller.java
```diff
@@ -49,7 +49,7 @@ public MatlabInstaller(String id) {
     }
 
     public String getRelease() {
-        return this.release.trim();
+        return this.release;
```
sameagen-MW, line 52: What cases caused the removal of the `trim` here?

sameagen-MW: Ah I see, it's just trimmed later.

sameagen-MW (approved):
Do we have any tests to cover these cases at the moment?

src/main/java/com/mathworks/ci/tools/MatlabInstaller.java
davidbuzinski: Yeah, mostly just that it seems weird for a getter to change the value of the property. We were already trimming later so I just removed it.

davidbuzinski:
Not sure what's going on with the flaky macos test

PR #400, opened 2025-01-22
Update release format for install automatically

davidbuzinski:
Change Automatic Installation name to use capital R in version name (ex: R2024b instead of r2024b).

### DIFF
--- a/src/main/java/com/mathworks/ci/tools/MatlabInstaller.java
+++ b/src/main/java/com/mathworks/ci/tools/MatlabInstaller.java
@@ -170,6 +170,10 @@ public class MatlabInstaller extends ToolInstaller {
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/mathworks/ci/tools/MatlabInstaller.java
```diff
@@ -170,6 +170,10 @@ private MatlabRelease parseRelease(String release) throws InstallationFailedExce
             }
         }
 
+        if (name.startsWith("r")) {
```
davidbuzinski, line 173: Honestly probably don't need the if statement if we're sure that the release format will always be "R<year>", but I figured it can't hurt to verify.

Vahila (approved):

